Issue #2, opened 2015-09-10
Bug: nuget package explorer appends api/v2/package to every url


304NotModified:
Copy of https://npe.codeplex.com/workitem/149

-------
Bug: nuget package explorer appends api/v2/package to every url

Full Thread Discussion: https://github.com/themotleyfool/Klondike/issues/34#issuecomment-43866074 

[quote] 
It looks like NuGet Package Explorer automatically appends api/v2/package to whatever URL you tell it to publish to. 

http://localhost.klondike.foolhq.com/ => http://localhost.klondike.foolhq.com/api/v2/package 

http://localhost.klondike.foolhq.com/api/packages => http://localhost.klondike.foolhq.com/api/pacakges/api/v2/package 

None of these URIs are valid in Klondike. The correct place to put packages is to /api/packages. 

I could put redirects into Klondike to work around this, but really the client should not be doing this. In my opinion this bug is in NuGet Package Explorer, not Klondike. 
[/quote] 

[quote] 
The behavior of their client is different from NuGet command line client, Visual Studio package manager, Package Manager Console, and even OctopusDeploy 
[/quote]

304NotModified:
This is correctly observed.

The code who does this can be found here: 
https://github.com/304NotModified/NuGetPackageExplorer/blob/master/PackageViewModel/GalleryServer.cs

PR #20, opened 2015-11-22
Add standard .gitattributes + normalize line-endings

caioproiete:
Commit #1 adds standard .gitattributes to normalize line-endings for all contributors moving forward
Commit #2 normalizes the line-endings of all text files to agree with the .gitattributes file added


304NotModified:
https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/pull/20/files?w=1

304NotModified:
I think we should exclude the folders which has always binaries? (eg lib)

caioproiete:
Hey Julian,

These assemblies are being referenced directly in the .csproj:
https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/blob/master/PackageExplorer/PackageExplorer.csproj#L104-L112

What do you have in mind? Replace these with NuGet package references?

I'm not familiar with these 3 libraries, but looks like there are NuGet packages for them:

* [GrayscaleEffect](https://www.nuget.org/packages/GrayscaleEffect/)
* [AvalonEdit](https://www.nuget.org/packages/AvalonEdit/)
* [Ookii.Dialogs](https://www.nuget.org/packages/Ookii.Dialogs/)

Do you or @dsplaisted know if these are the right ones?

/cc @304NotModified

---

ps: Not sure what is the relation with this PR though... Maybe open a separate issue/PR? I'm happy to do it.

304NotModified:
This has been merged. Thanks!

PR #25, opened 2016-01-14
update AboutWindow replace codeplex with github url

campersau:


304NotModified:
Thanks also for this one :)

Issue #22, opened 2015-11-27
"Open with" in Windows 10: The plugin that registers to read this file type fails to read the content of this file


304NotModified:
If I open an nupkg with double click. When starting NuGetPackageExplorer and then "open a local package", it works. 

> **\* The plugin that registers to read this file type fails to read the content of this file. ***

![image](https://cloud.githubusercontent.com/assets/5808377/11440836/2b582ae8-9506-11e5-847a-6aa4974ea4a9.png)

We should at least log the message? I think it could a permission/security issue. 

https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/blob/4ba823deb94a360cc95bdf4775349d5bd65822f7/PackageViewModel/Commands/ViewContentCommand.cs#L101

it should be: 
![image](https://cloud.githubusercontent.com/assets/5808377/11440921/fd1179c2-9506-11e5-8345-3832a7302473.png)

tested with 3.9 and 3.10 RC2

304NotModified:
Can't reproduce in 3.11. Close for now.

elv1s42:
Hello. 
I have the same problem with Windows 10 and Nuget Package Explorer 3.13.

> *** The plugin that registers to read this file type fails to read the content of this file. ***

> System.ArgumentNullException: Value cannot be null.
Parameter name: source
   at System.Linq.OrderedEnumerable\`2..ctor(IEnumerable\`1 source, Func\`2 keySelector, IComparer\`1 comparer, Boolean descending)
   at PackageExplorer.AssemblyFileViewer.GetView(String extension, Stream stream)
   at PackageExplorerViewModel.ViewContentCommand.ShowFile(PackageFile file)

304NotModified:
Hi!

Which file do you like to read (.dll etc)? Can you post a screenshot?

Thanks!

elv1s42:
Problem appears for .dll and .exe files, here is a screenshot:

![screenshot](https://cloud.githubusercontent.com/assets/6689884/14496807/207ff2a0-019d-11e6-8189-2ce127aa48fd.png)

.xml and .config files do not have this problem.
Issue appears only when I open an nupkg with double click. When starting NuGetPackageExplorer and then "open a local package", it works fine.

304NotModified:
Thanks! I will check this in the source code

304NotModified:
[...]

PR #31, opened 2016-01-24
Print exception for "plugin read failed"

304NotModified:
Can't reproduce https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/issues/22 with a local build. So printing the exception to the window in case of this error. 

image

### DIFF
--- a/PackageViewModel/Commands/ViewContentCommand.cs
+++ b/PackageViewModel/Commands/ViewContentCommand.cs
@@ -95,10 +95,10 @@ namespace PackageExplorerViewModel
                         }
                     }
                 }
-                catch (Exception)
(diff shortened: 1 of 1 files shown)


304NotModified:
@dsplaisted @rikoe 

Please your opinion on this change.

Printing the full exception isn't nice, but it will make the fix and maintenance of this tool a lot easier.

Another option is to log it.

dsplaisted:
Yes, I would display the full exception there. It's not recommended in general, but I think it makes more sense in an app targeting developers, especially if the exception is probably coming from a plugin.

304NotModified:
Thanks for the feedback!

PR #32, opened 2016-01-24
Add option for appending V2 API to URL

### LINKED ISSUE #2: Bug: nuget package explorer appends api/v2/package to every url

304NotModified:
Fixes #2 

image

@dsplaisted , @rikoe 

Please your thoughts on this change. Removing always `api/v2/package` would be a non-backwards-compatible change. (e.g. you need to change the publish URL when this is released).

The option is maybe not the best solution, but it's an improvement and makes the process more transparent.


304NotModified:
I think this is at least an improvement. So merge. We can change this in the future.

PR #35, opened 2016-01-24
Disable Windows Store popup

### LINKED ISSUE #33: Remove "Great! Your are running on Windows 8"
image

- First of all, Windows 8 isn't great...
- Second, I'm running Window 10
- Last, but not least, this version is not maintained and not good rated. Last release in 2013.  

Proposal: remove this popup, at least until the version in the Window Store is maintained. 

image

304NotModified:
It doesn't work well with Windows 10 and the Windows Store application
is not maintained.

image

fixes #33


304NotModified:
merged this already as it was really annoying in Windows 10

PR #48, opened 2016-02-22
Support of server authentication

304NotModified:
From
https://npe.codeplex.com/SourceControl/network/forks/jkalmbach/ProxySupport/contribution/8329

> Currently the Package Explorer uses no authetication to communicate with the nuget server. This works in most cases. If you work in a company which uses an own nuget-server which requres autehtication, then the Package Explorer dies not work, because it always returns "401 Unauthorized". One solution is not to just use the default credentials to allow authetntication with the current user.
This change has no side effects for anonymous access, this works as before. Now also the servers which require autehtication will work.


304NotModified:
superseded by #49

ben555:
+1. I get "An error occurred while processing this request.. The remote server returned status code 401."

304NotModified:
@ben555 does your case need a custom login form? (to enter the credentials)?

Issue #4, opened 2015-09-10
Windows Authentication fails


304NotModified:
Copy of https://npe.codeplex.com/workitem/142

-----
Hello folks, 
I recently tried out the tool NuGet Package Explorer, and I have to say that it is a very comfortable tool. 
On my NuGet server I activated the Windows Authentification in my IIS properties. 
The problem is that the Package Explorer responds me: 
Failed to process request: 'Unauthorized'.
How about adding Windows Authetification support? 

greetings 
Nico
comments

WhiteIntel wrote Apr 14, 2014 at 9:31 PM

Anyone?
WhiteIntel wrote May 7, 2014 at 9:56 PM

Does anyone take care about the posted issues?!
dobossandor wrote Sep 15, 2014 at 5:48 PM

Same issue with Basic authentication.
Nuget Package Explorer 3.9.0.0
dobossandor wrote Sep 15, 2014 at 5:56 PM

Something went wrong between 3.6.0.0 and 3.7.0.0.
Just tried these. 3.6.0.0 works, 3.7.0.0 isnt. (At least with basic auth)
rouslanp wrote May 13 at 5:50 PM

Windows Authentication is an extremely important feature for those who dared to maintain their own NuGet repositories. Please fix this bug!

--------

rikoe:
My merge in #16 should fix this - I just need to rebase and test my fix on this repo.

304NotModified:
:+1:

FYI 

I think rebase won't work, because the commit sha are different every import (from codeplex) 

If it does work, let me know. :)

304NotModified:
@rikoe did you see the Admin invite?

rikoe:
I didn't, I'll take a look. Thanks!

304NotModified:
Fixed by #49

Issue #16, opened 2015-11-22
Merge in


caioproiete:
([originally created](https://github.com/dsplaisted/NuGetPackageExplorer/issues/4) by @304NotModified)

Proposal to merge from other forks:

- [x] Support for pushing to authenticated NuGet feed
- [x] Include default credentials in web requests
- [x] Fallback to Packages() service for all versions

304NotModified:
@rikoe can you check this?

rikoe:
Will do soon. Thanks for reminding me.

304NotModified:
OK Cool :)

Like to release it after it.

304NotModified:
@rikoe any idea when you have time for this? 

Thanks in advance!

rikoe:
Hi - sorry, we've had a baby recently so struggling to find time. Will try to do it by Wednesday.

304NotModified:
@rikoe any success with finding some time?

rikoe:
Hi @304NotModified I have just started pulling the latest code to try and get this done, and have just seen that you have created some branches of your own for this (based on my implementation, or your own, I haven't checked)

I can't really complain  about this, as I did not respond to your previous message and it has been a while since I last said I would try to find some time.

Would you like to continue on on your own, or should I create a PR for these features based on my earlier work?

304NotModified:
A PR would be great!

I was indeed working on this, but I have difficulties to test it as I don't have the environment to test it.

rikoe:
Testing is the real problem, as I don't have a similar environment with Windows Authentication as when I set this up any more. I have submitted the PR but it needs to be tested so maybe leave it open. I have to see whether I can set up a VM to do it or maybe test it at work.

Anyway, this is all I could manage for today :smile:

304NotModified:
Fixed by #49

PR #49, opened 2016-02-22
Fix credentials + fallback for legacy API

rikoe:
I created these changes in 2014 when I was trying to fix NuGetPackageExplorer to work with a local Active Directory install of ProGet.

The changes were both to be able to download packages from and push packages to the server using Windows Integrated Authentication, as supported by ProGet.

Please review and approve/discard as you see fit - hopefully we can address #16 and #4 with this.


304NotModified:
:)  thanks! 

Will try to release this week a new version.

sbochoa:
@rikoe Hello thanks for this change but what about if im not in the domain is there a chance to popup a login dialog to enter the credentials just like visual studio does?

rikoe:
Well, you'd need to be clear about your use case. When would you use
NugetPackageExplorer and not be in the domain already?

My fix was simply for authenticating against the NuGet server with the same
credentials you are already logged in with. It's just a few lines of code
to enable capturing the user's current credentials.

What you're asking for sounds a lot more complex, effectively supporting
multiple modes of authentication. Also keep in mind that the Visual Studio
dialog you are referring to is actually quite a lot more complex than it
first appears. It is hosting a browser control that authenticates against a
hosted OAuth security provider as far as I know (if we are referring to the
same one).

If you really need what you are asking for I would suggest opening a
separate issue for it describing exactly what you are authenticating
against and how (maybe with screenshots).

We can then try and prioritise it, but if it's not a common use case you
may need to fork it and try yourself! :-)

304NotModified:
@rikoe is this PR better than https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/pull/48?

rikoe:
It is a superset of #48 - it includes the same change to PackageDownloader.cs, but adds default credentials in a bunch of other places.

The one additional thing it does, is that (at the time) ProGet used a legacy NuGet API for searching packages by id, which didn't work with NuGetPackageExplorer, so I added some code to fall back to the legacy API

304NotModified:
Fixes https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/issues/4

WiteCastle:
How would one authenticate with a VSO nuget feed? https://blogs.msdn.microsoft.com/visualstudioalm/2015/08/27/announcing-package-management-support-for-vsotfs/

304NotModified:
Hi @WiteCastle , only Windows authentication (without login form) is currently supported. 

https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/issues/92

PR #72, opened 2016-03-01
Added support for using Portable Dialog

ravensorb:
Inital version PortableDialog to use NuGet Client framework and to use MVVM pattern
Added support for Xamarin in PortableDIalog


ravensorb:
This should build -- is it possible that nuget on appveyor cannot download the Nuget Client v3 packages?

PackageExplorer/PortableLibraryDialog.xaml.cs
```diff
@@ -53,9 +39,9 @@ private void CancelButton_Click(object sender, RoutedEventArgs e)
 
         private void EvaluateButtonEnabledState(object sender, RoutedEventArgs e)
         {
-            var _allCheckBoxes = new CheckBox[] { NetCheckBox, SilverlightCheckBox, WindowsCheckBox, WPCheckBox };
-            var count = _allCheckBoxes.Count(p => p.IsChecked == true);
-            OKButton.IsEnabled = count >= 2;
+            //var _allCheckBoxes = new CheckBox[] { NetCheckBox, SilverlightCheckBox, WindowsCheckBox, WPCheckBox, XamarinAndroid, XamariniOS };
```
304NotModified: please remove commented code.

304NotModified:
Thanks @ravensorb !

Too bad we cannot accept this with the current file headers. We don't like the possible copyright issues. Proposal: remove the headers, your are of course the author of the commits.

If that is possible, I will try to fix AppVeyor. Probable needs some config.


PS: compare without newline changes: https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/pull/72/files?w=1

ravensorb:
Ok, I found out that the Nuget package I referenced was actually an old one :) So I updated that, removed the headers, removed the commented code, and did some more clean up.  The only thing outstanding is I am not 100% sure yet how to use the NuGet framework library to VALIDATE the selected platforms so I am sticking with a simple option right now.

304NotModified:
OK thanks! Will look at the code in-depth

PackageExplorer/Models/NuGetFrameworkModel.cs
```diff
@@ -0,0 +1,255 @@
...
+		/// <value>The selected frameworks.</value>
+		public IEnumerable<NuGetFramework> SelectedFrameworks
+		{
+			get
+			{
+				var selectedPlatforms = (from platform in PortableFrameworks
+										 where platform.IsSelected
+										 select platform.SelectedItem?.Framework ?? platform?.Framework);
+
+				return selectedPlatforms;
```
304NotModified: On other code parts Linq Extensions are used instead of the query syntax. I also prefer that. Can you rewrite that to 

`PortableFrameworks.Where(...).Select(..)` ? (Reshaper can do that for your)

PackageExplorer/Models/NuGetFrameworkModel.cs
```diff
@@ -0,0 +1,255 @@
...
+
+			InitializePCLModel();
+		}
+
+		/// <summary>
+		/// Initializes the PCL model.
+		/// </summary>
+		private void InitializePCLModel()
+		{
+			var frameWork = new NuGetTargetFrameworkItem { Name = ".NET Framework" };
```
304NotModified: There a lot of code duplication here. Can we remove that? I was thinking of a dictionary which contains (at least) tne name & shortname

PackageExplorer/Models/NuGetFrameworkModel.cs
```diff
@@ -0,0 +1,255 @@
...
+			frameWork = new NuGetTargetFrameworkItem { Name = "Xamarin For Android", ShortName = "MonoAndroid", Framework = Frameworks.FindByIdentifierShortName("MonoAndroid").FirstOrDefault() };
+			PortableFrameworks.Add(frameWork);
+		}
+	}
+
+	/// <summary>
+	/// Class NuGetTargetFrameworkItem.
+	/// </summary>
+	[PropertyChanged.ImplementPropertyChanged]
+	public class NuGetTargetFrameworkItem
```
304NotModified: Please move to separate file

PackageExplorer/Models/NuGetFrameworkModel.cs
```diff
@@ -0,0 +1,255 @@
...
+	{
+		/// <summary>
+		/// Gets or sets the name.
+		/// </summary>
+		/// <value>The name.</value>
+		public string Name { get; set; }
+
+		/// <summary>
+		/// Gets or sets the short name.
```
304NotModified: Thanks for all the docs, but this won't help much. What is a short name? (maybe add an example?)

PackageExplorer/Models/NuGetFrameworkModel.cs
```diff
@@ -0,0 +1,255 @@
...
+	public static class NuGetTargetFrameworkItemExtensions
+	{
+		/// <summary>
+		/// Ases the targeted platform.
+		/// </summary>
+		/// <param name="items">The items.</param>
+		/// <returns>System.String.</returns>
+		public static string AsTargetedPlatformPath(this ObservableCollection<NuGetTargetFrameworkItem> items)
+		{
+			var selectedPlatforms = (from platform in items
```
304NotModified: This is duplication code. Please remove it

ravensorb: Not sure I understand -- there are two extension methods based on which type of collection is being used.  Where is the duplication?

PackageExplorer/Models/NuGetFrameworkModel.cs
ravensorb: Done

PackageExplorer/Models/NuGetFrameworkModel.cs
ravensorb: Not sure I follow -- the entire point of this set of extensions is to leverage the existing objects within the NuGet class libraries.  These extensions just make it a lot easier to work with for the kinds of operations needed for the UI.

PackageExplorer/Models/NuGetFrameworkModel.cs
ravensorb: Done

ravensorb:
Just updated based on you feedback -- thanks!

PackageExplorer/Models/NuGetFrameworkModel.cs
304NotModified:
```
frameWork.Items.Add(new NuGetTargetFrameworkItem { Name = ".NET Framework 4 and higher", ShortName = "net4", Framework = Frameworks.FindByIdentifierShortName("net40").FirstOrDefault() });
```

is duplicate of
```
frameWork.Items.Add(new NuGetTargetFrameworkItem { Name = ".NET Framework 4.0.3 and higher", ShortName = "net403", Framework = Frameworks.FindByIdentifierShortName("net403").FirstOrDefault() });
```

etc

The non-duplicate things are  ".NET Framework 4.0.3 and higher" (Name) and "net403" (ShortName)

PackageExplorer/Utilities/NuGetExtensions.cs
```diff
@@ -0,0 +1,271 @@
...
+		/// <summary>
+		/// Finds the name of the by framework.
+		/// </summary>
+		/// <param name="frameworkMapping">The framework mapping.</param>
+		/// <param name="frameworkName">Name of the framework.</param>
+		/// <returns>NuGetFramework.</returns>
+		public static NuGetFramework FindByFrameworkName(this IFrameworkMappings frameworkMapping, string frameworkName)
+		{
+			var result = frameworkMapping.EquivalentFrameworks.FirstOrDefault(x =>
+				string.Compare(x.Key.Framework, frameworkName, StringComparison.InvariantCultureIgnoreCase) == 0 ||
```
304NotModified: Why are we using compare, and not equals?

PackageExplorer/Models/NuGetFrameworkModel.cs
ravensorb: They are actually different when it comes to portable framework identifiers and paths.

PackageExplorer/Utilities/NuGetExtensions.cs
```diff
@@ -0,0 +1,271 @@
...
+
+			return result.Value;
+		}
+
+		/// <summary>
+		/// Finds the short name of the by short name identifier.  Example: Searching for sl4 would return the framework for SilverLight, Version=4.0
+		/// </summary>
+		/// <param name="frameworks">The frameworks.</param>
+		/// <param name="identifierShortName">Short name of the identifier.</param>
+		/// <returns>IEnumerable&lt;NuGetFramework&gt;.</returns>
```
304NotModified: You can use the `<see` tag instead of `IEnumerable&lt;NuGetFramework&gt;.` here.

But in this Case these `<returns>` won't add anything, so please remove it.

PackageExplorer/packages.config
```diff
@@ -0,0 +1,7 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="Fody" version="1.29.4" targetFramework="net45" developmentDependency="true" />
```
304NotModified: What's Fody?

PackageExplorer/Utilities/NuGetExtensions.cs
ravensorb: equals is case and locale sensitive -- Compares is more of a "best practice" from a development perspective.  Since the NuGet packages have the potential to be localized, I didn't want anything to break :)

304NotModified: > equals is case and locale sensitive

That's not the case when passing ` StringComparison.InvariantCultureIgnoreCase` to Equals?

PackageExplorer/packages.config
ravensorb: [Fody](https://github.com/Fody/Fody) is a tool that makes certain things easier :smile: when it comes to getting right of "boiler plate" code.  I use it to handle the code for [PropertChanged](https://github.com/Fody/PropertyChanged) notifications a lot as the code for that is almsot 100% redundant in all cases.

PackageExplorer/Utilities/NuGetExtensions.cs
ravensorb: I suppose however it assumes that the object in question is not null - where compare handles null on either operand.  I guess I could switch to something like 
`x.Key.Framework?.Equals(frameworkName, StringComparison.InvariantCulterIgnoreCase)`
Just not sure what the really gets us?

PackageExplorer/Utilities/NuGetExtensions.cs
ravensorb: I usually keep my comments in a specific format as it makes generate documentation more consistent.  Also the "see" tag doesn't always work correctly on IEnumerable return results as most of the time it takes you to the IEnumerable documetnation rather than the object that is enumerable.

PackageExplorer/packages.config
304NotModified: Cool! will read into it

PackageExplorer/Utilities/NuGetExtensions.cs
304NotModified: Well do you agree that it won't add anything now? (for the end-user)

ravensorb: I updated the comment in the latest push -- should have more info in almost all of the returns comments to add more value.  Will that be ok?

ravensorb:
Everything ok to merge this into the main branch?

PackageExplorer/Models/NuGetFrameworkModel.cs
```diff
@@ -0,0 +1,123 @@
...
+
+			frameWork = new NuGetTargetFrameworkItem { Name = ".NET for Windows Store apps", ShortName = "win8", Framework = Frameworks.FindByIdentifierShortName("win8").FirstOrDefault() };
+			PortableFrameworks.Add(frameWork);
+
+			frameWork = new NuGetTargetFrameworkItem { Name = "Xamarin For iOS", ShortName = "xamarinios", Framework = Frameworks.FindByIdentifierShortName("xamarinios").FirstOrDefault() };
+			PortableFrameworks.Add(frameWork);
+
+			frameWork = new NuGetTargetFrameworkItem { Name = "Xamarin For Android", ShortName = "MonoAndroid", Framework = Frameworks.FindByIdentifierShortName("MonoAndroid").FirstOrDefault() };
+			PortableFrameworks.Add(frameWork);
+		}
```
304NotModified: Please try to remove the code- / config duplication in this method. My previous comment on this issue has been lost I see.

PackageExplorer/Utilities/NuGetExtensions.cs
```diff
@@ -0,0 +1,271 @@
...
+		/// <param name="frameworkIdentifier">The framework identifier.</param>
+		/// <param name="includeEquivalentFrameworks">if set to <c>true</c> [include equivalent frameworks].</param>
+		/// <returns>A collection of NuGetFramework items that match the framework specified.</returns>
+		public static IEnumerable<NuGetFramework> FindByIdentifierShortName(this IPortableFrameworkMappings frameworkMapping,
+			string frameworkIdentifier, bool includeEquivalentFrameworks = false)
+		{
+			var frameworkName =
+				DefaultFrameworkMappings.Instance.IdentifierShortNames.FirstOrDefault(
+					x =>
+						string.Compare(x.Value, frameworkIdentifier, StringComparison.InvariantCultureIgnoreCase) == 0 ||
```
304NotModified: please replace `.Compare` with `.Equals`.

PackageExplorer/Utilities/NuGetExtensions.cs
```diff
@@ -0,0 +1,271 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace NuGet.Frameworks
+{
+	/// <summary>
+	/// Class for Extensions related to NuGetFramework objects.
+	/// </summary>
+	public static class NuGetFrameworkExtensions
```
304NotModified: Can you add some examples (`<example`) on this methods? I have difficulties to follow the logic here

PackageExplorer/ViewModels/PortableLibraryViewModel.cs
```diff
@@ -0,0 +1,50 @@
+﻿using System.Windows.Input;
+using NuGet.Frameworks;
+
+namespace PackageExplorer.ViewModels
+{
+	public class PortableLibraryViewModel
```
304NotModified: Please a XML comments why this class is needed.

PackageExplorer/Wpf/NuGetTargetFrameworkItemDataTemplateSelector.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Windows;
+using System.Windows.Controls;
+using NuGet.Frameworks;
+
+namespace PackageExplorer.Wpf
+{
+	/// <summary>
+	/// Class PortableClassDataTemplateSelector.
+	/// </summary>
+	public class NuGetTargetFrameworkItemDataTemplateSelector : DataTemplateSelector
```
304NotModified: Please add some valuable comments.

304NotModified:
Hi!

I checked the code and some parts are very difficult to follow (and review). I'm sure this isn't rocket science, so need to improve that otherwise it's too difficult to maintain this code.

PackageExplorer/Models/NuGetFrameworkModel.cs
ravensorb: I am not sure I understand, there is nothing in here that is duplicate.  This is the core of the most recent enhancement, it moves ALL of the target platforms into a single are and leverages the NuGet libraries. Everything in here is needed to support the different platforms within the selection dialog.  Eventually I'd love to move this into the NuGet libraries itself (or at least something similar to this code).

PackageExplorer/Utilities/NuGetExtensions.cs
ravensorb: I'll add more comments -- these were refactored into extensions specifically because the current NuGet framework is very convoluted in how it organizes its internal data structures.

PackageExplorer/Utilities/NuGetExtensions.cs
ravensorb: They are not the same and have the potential for null value exceptions.  Adding a null check would be just as complicated --- unless you want me to switch to using the ?. in C# 6?

PackageExplorer/ViewModels/PortableLibraryViewModel.cs
ravensorb: This class is part of the MVVM pattern that allowed for separation of the code from the UI.  I'll add some comments to this effect.

PackageExplorer/Models/NuGetFrameworkModel.cs
304NotModified: See the following examples:

These calls:

```
frameWork.Items.Add(new NuGetTargetFrameworkItem { Name = "YYYY", ShortName = "XXXXX", Framework = Frameworks.FindByIdentifierShortName("XXXXX").FirstOrDefault() });
PortableFrameworks.Add(frameWork);
```

are duplicate (code duplication)

The identifier "xamarinios" etc. is duplicate (config duplication):

```
frameWork = new NuGetTargetFrameworkItem { Name = "Xamarin For iOS", ShortName = "xamarinios", Framework = Frameworks.FindByIdentifierShortName("xamarinios").FirstOrDefault() };
```

PackageExplorer/ViewModels/PortableLibraryViewModel.cs
ravensorb: Done

PackageExplorer/Wpf/NuGetTargetFrameworkItemDataTemplateSelector.cs
ravensorb: Done - also added a "see" comment to related it to where it being used

ravensorb:
Would you like me to pull the extension methods out in to into a separate nuget package?

304NotModified:
What do you mean?

ravensorb:
the static class NuGetFrameworkExtensions -- If you have concerns about that being too complicated for the solution, I would be more than willing to create a new NuGet package (That I maintain) that contains those extensions.  Then NuGetPackageExplorer could just reference that nuget package rather than have the classes in your branch.  Just a thought :)  Otherwise, I think I updated everything else based on your feedback.

304NotModified:
@ravensorb 

> the static class NuGetFrameworkExtensions -- If you have concerns about that being too complicated for the solution, I would be more than willing to create a new NuGet package (That I maintain) that contains those extensions.

That's a funny idea. But isn't the case that is too complicated for me, but it's too complicated for it's purpose. I'm pretty sure you can refactor the stuff to get some improved code quality.

ravensorb:
You are welcome to try -- that was the best I could come up with given the nature of the Nuget framework code :smile: They have things split into multiple different collections with only a "loose" relationship (its almost like they took a normalized database and turned it into a class structure).

304NotModified:
> You are welcome to try

OK, but if you can fix first that it builds on AppVeyor ;) 

I think you need to enabled nuget restore for the project.

304NotModified:
@ravensorb can you fix the build?

ravensorb:
Sure -- can you give me access to the appveryor build env?

304NotModified:
I think it's easier that you will test your fork in appveyor (set up appveyor is done by appveyor.yml)

You sign in on appveyor (with your github account) and add you fork :) ("add project"). If you need some help on that, just let me know.

ravensorb:
if you add a "nuget restore" in the before build, it should work

https://www.appveyor.com/blog/2014/03/18/about-nuget-package-restore

304NotModified:
yes, so please add it to appveyor.yml and add in the PR :)

PackageExplorer/Models/NuGetFrameworkModel.cs
ravensorb: Again, they are not duplicates :) Each line has a small different to handle the variations in the framework definitions.

ravensorb:
Just added the support for package restore in the appveyor file

304NotModified:
Can you fix the merge conflict? Hopefully it builds then

304NotModified:
I have tested it, works good.

image


The merge conflict is easy to solve.

The code duplication is a todo.

304NotModified:
fixed the nuget issue. Ancient nuget.exe in the root (v1.5...)

304NotModified:
superseded by https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/pull/138

PR #78, opened 2016-03-22
Add support for exclude attribute on dependency

onovotny:
The latest .NET Core packages seem to have added a new attribute on the dependency. This PR adds support and fixes the crash without it.

Here's the example package:

Feed: https://dotnet.myget.org/F/dotnet-core
Package: System.Reflection.TypeExtensions rc3-23922

Try loading that without this and NPE crashes. With this, support is added throughout the UI for it.


304NotModified:
Cool! Thanks @onovotny !

304NotModified:
This has been released! (3.13)

PR #80, opened 2016-03-29
Resolved OutOfMemory exceptions for multiple opening of packages >256…

iHateInventNames:
Resolved OutOfMemory exceptions for multiple opening of packages >256Mb unpacked (tested on Wt)
Also, looks like it improves responsivness while opening big packages.

### DIFF
--- a/Core/Packages/ZipPackageFile.cs
+++ b/Core/Packages/ZipPackageFile.cs
@@ -8,19 +8,13 @@ namespace NuGet
 {
     internal class ZipPackageFile : PackageFileBase
     {
-        private readonly Func<MemoryStream> _streamFactory;
+        private readonly Func<Stream> _streamFactory;
 
         public ZipPackageFile(PackagePart part) 
             : base(UriUtility.GetPath(part.Uri))
         {
             Debug.Assert(part != null, "part should not be null");
-
-            byte[] buffer;
-            using (Stream partStream = part.GetStream())
-            {
-                buffer = partStream.ReadAllBytes();
-            }
-            _streamFactory = () => new MemoryStream(buffer);
+            _streamFactory = () => part.GetStream();
         }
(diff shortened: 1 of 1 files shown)


304NotModified:
Thanks! Will test it!

Core/Packages/ZipPackageFile.cs
```diff
@@ -8,19 +8,13 @@ namespace NuGet
...
         {
             Debug.Assert(part != null, "part should not be null");
-
-            byte[] buffer;
-            using (Stream partStream = part.GetStream())
-            {
-                buffer = partStream.ReadAllBytes();
-            }
-            _streamFactory = () => new MemoryStream(buffer);
+            _streamFactory = () => part.GetStream();
```
304NotModified: what the benefit of the lambda here?

304NotModified:
there seems to be an issue:

1. open file from online feed
2. save as


image

304NotModified:
I think you need stream.CopyTo. I think that one is also  more efficient and won't result in these errors

iHateInventNames:
Looks like the stream was opened for write before and was not properly closed?

304NotModified:
I guess the stream is readonly and only writable after copying. Is there a big package on nuget.org so I can test things?

Will close this PR as this leads to issues, but a new PR is of course fine :)

iHateInventNames:
Yes, there is [wt](https://www.nuget.org/packages/Wt/3.3.5-rc1) on nuget.org

304NotModified:
Thanks! I saw the issue with a package of +1000MB. Tried to fixed it, but failed.

iHateInventNames:
[Fixed](https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/pull/87)

PR #81, opened 2016-03-30
Show package published date in package chooser

campersau:
So you can see if a package is actively maintained or not.


304NotModified:
Nice idea!

PackageExplorer/Converters/DateTimeOffsetConverter.cs
```diff
@@ -0,0 +1,27 @@
...
+            if (value is DateTimeOffset)
+            {
+                return ((DateTimeOffset)value).DateTime.ToLocalTime().ToString();
+            }
+            return null;
+        }
+
+        public object ConvertBack(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            throw new NotImplementedException();
```
304NotModified: Should this be a notsupportedException?

campersau: A lot of the other converters also throw a `NotImplementedException`. See:
- https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/blob/master/PackageExplorer/Converters/AndLogicConverter.cs#L25
- https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/blob/master/PackageExplorer/Converters/BooleanToStringConverter.cs#L19
- https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/blob/master/PackageExplorer/Converters/BooleanToVisibilityConverter.cs#L41

...

But if you want I can change it.

304NotModified:
Good point. Thanks!

304NotModified:
I get `datetime.min` dates for hidden packages. Would be nice to fix that :)

Also, would a short date (e.g. 31 march 2016) not better?

image

image

campersau:
Updated to use `ToShortDateString` instead and hide the `Published` date for `Unpublished` packages.

---

The commit https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/pull/81/commits/7e3c4ab83e4655ae456061def1de40c1692d72f2 also tries to fix a timezone issue. I am in germany and the unlisted packages weren't displayed in italic because [`IsUnlisted`](https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/blob/master/Core/Packages/PackageInfo.cs#L53) always returned `false`, because the UtcOffset was different from the parsed DateTimeOffset. ([DateTimeOffsets are compared using their UtcDateTime.](http://referencesource.microsoft.com/#mscorlib/system/datetimeoffset.cs,441))
The XsdDateTime parser also uses this logic: http://referencesource.microsoft.com/#System.Xml/System/Xml/Schema/XsdDateTime.cs,465

image

It would be nice if someone could verify that this behaviour works in other time zones too.

304NotModified:
I'm in UTC+2 (or UTC+1 and daylight saving?) and it works very nice :)



> Also, would a short date (e.g. 31 march 2016) not better?

I think I meant long date... :fearful:  What do you think? Isn't the long date ("Wednesday, May 16, 2001") not better? (vs "5/16/2001")? It's far more readable, but it's a bit (too) long maybe. NuGet.org also uses the long date.

dsplaisted:
We could also consider using [Humanizer](http://humanizr.net/) to get a description like "2 weeks ago" for the date.

304NotModified:
That's also nice, but then I really like a tooltip with the long date 

(as I'm lazy and don't like to think which date it was, 2 weeks ago ;)

campersau:
Argh, just pushed a commit to use `ToLongDateString` :smile: 

image

In german the text is quite long so I increased the width to 200.

304NotModified:
Agile :innocent:

campersau:
Here are some screenshots with Humanizer:
image

image

So if we go for this option we definitely need tooltips with the absolute dates.

@304NotModified Personally I like number only dates the most in tables (e.g. `yyyy-MM-dd`), because they have always the same size.

Also Nuget uses two different formats. `yyyy-MM-dd` for the package list and `ToLongDateString` for the package versions list.

304NotModified:
Looks nice! I prefer humanizer with tooltip (long date)

campersau:
Updated to use Humanizer and Tooltip.
I have also included some Humanizer locale resources.

304NotModified:
Cool! Thanks! Will test it soon

304NotModified:
Really nice!

304NotModified:
:| 

Clickonce won't work again:

image

304NotModified:
ClickOnce issue fixed by removing resource files. (kinda hack), too bad I isn't possible on CodePlex to move back to a previous ClickOnce release :-1: 

Anyway, released. We could try the resources of Humaziner another time, although nothing of NPE is currently localized.

PR #82, opened 2016-04-13
Show "strong name" in assembly properties

304NotModified:
Show "strong name" in assembly properties + refactor AssemblyMetaData

If a strong name is good or not, is a discussion. But making it clear is a big win.

Example:
image

### DIFF
--- a/CodeExecutor/AppDomainWorker.cs
+++ b/CodeExecutor/AppDomainWorker.cs
@@ -36,9 +36,9 @@ namespace CodeExecutor
             return null;
         }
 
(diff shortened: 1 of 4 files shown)


dsplaisted:
If the full name has public key token, then the assembly is strong named. Do you think it's helpful to also show it separately?

304NotModified:
Yes, as there are a lot of users who don't know that (in my experience). There is a lot of confusion regarding strong naming, so I try to help with that.

But let me know if you think this is nonsense. ;)

304NotModified:
@dsplaisted Is it OK to merge and release this?

dsplaisted:
Sure, go ahead!

PR #96, opened 2016-04-25
Remove wrong sort by version which compared strings

campersau:
this is also not needed because the packages were already [sorted by published date](https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/blob/master/PackageViewModel/PackageChooser/PackageInfoViewModel.cs#L178)

Before:
image

After:
image

### DIFF
--- a/PackageViewModel/PackageChooser/PackageInfoViewModel.cs
+++ b/PackageViewModel/PackageChooser/PackageInfoViewModel.cs
@@ -294,9 +294,6 @@ namespace PackageExplorerViewModel
                 results = await Task.Run((Func<PackageInfo[]>)query.ToArray, token);
             }
 
-            // sort by Version descending
-            Array.Sort(results, (a, b) => b.Version.CompareTo(a.Version));
-
             return results;
         }
 


campersau:
This was introduced with https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/commit/6ec767632e2e9a80f2693fa6b4e6c38e77dd3265 (https://npe.codeplex.com/SourceControl/changeset/6ec767632e2e9a80f2693fa6b4e6c38e77dd3265)

The strange thing is that the [order by published date](https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/blob/6ec767632e2e9a80f2693fa6b4e6c38e77dd3265/PackageViewModel/PackageChooser/PackageInfoViewModel.cs#L130)  was already there. So maybe we should change that?

304NotModified:
Thanks!

304NotModified:
I'm not sure if this is a good change.

current:

image

your proposal:

image



I think it would be nice if we could click on the "version"  and "published" header to sort it (and that it would behave well)

campersau:
Yes, I know it can be improved (currently sorting on the detail grid is not implemented at all). But I think that the current behaviour is just wrong for version parts > 9. You can see it in your example that `4.4.0-beta11` is not displayed before `4.4.0-beta9`.

304NotModified:
The sorting is semver compatible. Also on nuget.org beta11 is after beta6. (which isn't a bug). See semver.org

campersau:
I forgot to say that sorting direction is *descending* by publish date. (That's why 11 is before 9.)
Also I think the sorting on nuget is more like mine proposal: https://www.nuget.org/packages/NLog/ or did I miss something?

304NotModified:
I think you're right about the latter. Anyway, I think it's debatable if it's an improvement.  If we make it optional (eg. by clicking on the header), then it would be a improvement for sure.

304NotModified:
Thanks for the PR, but IMO this isn't an improvement. The current sorting is more in line with nuget.org and semver. I cannot merge this, if it isn't optional.

304NotModified:
Thanks for this PR, but as already explained, it's an opinion if this is an improvement. Therefor we will not merge this PR. If you think we're wrong, just let us know.

PR #102, opened 2016-05-19
Add dependency on nu get.core library

BreeeZe:
In order to keep the core code up to date to the original NuGet code, I added a dependency on the NuGet.Core library : https://www.nuget.org/packages/NuGet.Core/
To make this possible I had to change the namespace from NuGet to NuGetPE and then remove duplicate code which is now sourced from the library.
It was not possible to phase out all duplicated code yet, but Semver and all framework identifiers are now used from the library package.

Hopefully this is useful to you, at least it is to me, since it is now easier to keep up with stuff like the new .Net Core framework identifiers


304NotModified:
Thanks! Will check this out, but could take some weeks for me

BreeeZe:
Sure,
And I got the appveyor build with nuget restore working too :)
For some reason the described method of just adding "nuget restore" to the appveyor.yml file didn't work.. (https://www.appveyor.com/blog/2014/03/18/about-nuget-package-restore)
It gave an "unknown command: 'restore'" error...
So I had to resort to adding some powershell commands.

304NotModified:
that's wierd as `nuget restore` always worked for me

304NotModified:
don't we need to use nuget.client?

BreeeZe:
If I remember correctly NuGet changed some classes and method parameters compared to the ones in NuGetPackageExplorer, so I could not substitute the duplicate versions easily with the ones from the NuGet.Core package.
At the time (oct 2015) I changed what I could with the time I had for this (I am using NuGetPackageExplorer at work)
But I'll see if I can invest some time in the coming weeks to look into this again (either privately or at work), would be nice to finish te job :)

304NotModified:
I like to merge this, but not sure if we like 152 changed files...

304NotModified:
superseded by https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/pull/147

(same stuff, included package update and conflict fix)

Issue #101, opened 2016-05-03
dateTimeOffset.LocalDateTime.ToLongDateString() Error


shuyuea3:
Version 3.15.0.0
namespace PackageExplorer

```
        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
        {
            if (value is DateTimeOffset)
            {
                DateTimeOffset dateTimeOffset = (DateTimeOffset)value;

                if (dateTimeOffset != Constants.Unpublished)
                {
                    //Use return dateTimeOffset.LocalDateTime.ToLongDateString(); Error Will automatically exit
                    return dateTimeOffset.LocalDateTime.ToShortDateString();//Use normal
                }
            }
            return null;
        }`

Yuexs:
I also encountered this problem

304NotModified:
Is there somewhere an exception /stacktrace to see? How could we reproduce this?

304NotModified:
Please use the template.  We are missing needed info now.

Yuexs:
I'm sorry, again according to the template

ksciana:
**Full stacktrace for this error:**
Aplikacja: NuGetPackageExplorer.exe
Wersja architektury: v4.0.30319
Opis: proces został przerwany z powodu nieobsłużonego wyjątku.
Informacje o wyjątku: System.ArgumentNullException
   w System.String.FormatHelper(System.IFormatProvider, System.String, System.ParamsArray)
   w System.String.Format(System.String, System.Object[])
   w Humanizer.Localisation.Formatters.DefaultFormatter.Format(System.String, Int32)
   w Humanizer.Localisation.Formatters.DefaultFormatter.DateHumanize(Humanizer.Localisation.TimeUnit, Humanizer.Localisation.Tense, Int32)
   w Humanizer.DateTimeHumanizeStrategy.DateTimeHumanizeAlgorithms.DefaultHumanize(System.DateTime, System.DateTime, System.Globalization.CultureInfo)
   w Humanizer.DateTimeHumanizeStrategy.DefaultDateTimeHumanizeStrategy.Humanize(System.DateTime, System.DateTime, System.Globalization.CultureInfo)
   w Humanizer.DateHumanizeExtensions.Humanize(System.DateTime, Boolean, System.Nullable`1<System.DateTime>, System.Globalization.CultureInfo)
[...]

PR #103, opened 2016-05-27
Fix error #101. System.ArgumentNullException in Humanize datetimeoffset.

ksciana:
Works on my machine ;)


304NotModified:
fixes #101

304NotModified:
Although I cannot reproduce #101, I think this works well. :)

304NotModified:
From tests it's returning the same values as `dateTimeOffset.LocalDateTime.Humanize();`, so no behavior change  :+1:

304NotModified:
Releasing! Thanks @ksciana !

PR #124, opened 2016-08-23
Pass false to utcDate parameter when humanizing LocalDateTime

### LINKED ISSUE #123: Incorrect humanized Published time in package chooser.

**Type** (choose one):

- Bug

**NPE version**: (e.g. 3.12) (see Help -> about)
Latest

**Installed from**: ClickOnce 

In case of a BUG:

- What is the current result?

In the package chooser dialog, the humanized published time is incorrect.  For example, I am at GMT+7 timezone,  a package published to NuGet.org one hour ago shows "8 hours ago"

It should show "1 hour ago"

The signature of Humanize() used in the converter:
```csharp
       public static string Humanize(this DateTime input, bool utcDate = true, DateTime? dateToCompareAgainst = default(DateTime?), CultureInfo culture = null);
 ```

In https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/blob/master/PackageExplorer/Converters/DateTimeOffsetHumanizeConverter.cs#L21,  it looks like that the call should have passed `false` to the utcDate parameter.

```csharp
                    return dateTimeOffset.LocalDateTime.Humanize(true, null, culture);
```

- Are there any work arrounds? no

I was checking Serilog.Sinks.Logentries.  However, as time flies the humanized string might be different in the future.

jeremymeng:
Fixes #123

### DIFF
--- a/PackageExplorer/Converters/DateTimeOffsetHumanizeConverter.cs
+++ b/PackageExplorer/Converters/DateTimeOffsetHumanizeConverter.cs
@@ -18,7 +18,7 @@ namespace PackageExplorer
 
                 if (dateTimeOffset != Constants.Unpublished)
                 {
-                    return dateTimeOffset.LocalDateTime.Humanize(true, null, culture);
+                    return dateTimeOffset.LocalDateTime.Humanize(false, null, culture);
                 }
             }   
             return null;


304NotModified:
tested and confirmed, thanks!

Sorry for the delay!

Issue #127, opened 2016-09-04
"The process cannot access the file because it is being used by another process." in MachineCache.ClearCache


svick:
**Type**: Bug

**NPE version**: 3.18.0.0, master

**Installed from**: ClickOnce

Opening a package from NuGet while another package (Marten 1.0.0-alpha-710, but I don't think that's relevant) is open in another instance of NPE throws:

```
System.IO.IOException: The process cannot access the file 'Marten.1.0.0-alpha-710.nupkg' because it is being used by another process.
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.FileInfo.Delete()
   at NuGet.MachineCache.ClearCache(DirectoryInfo cacheDirectory, Int32 threshold) in E:\Users\Svick\git\NuGetPackageExplorer\Core\Repositories\MachineCache.cs:line 104
   at NuGet.MachineCache.AddPackage(IPackage package) in E:\Users\Svick\git\NuGetPackageExplorer\Core\Repositories\MachineCache.cs:line 76
```

PR #128, opened 2016-09-04
Catch correct exception thrown from Delete()

svick:
Fixes https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/issues/127.

[The documentation for `FileInfo.Delete()`](https://msdn.microsoft.com/en-us/library/system.io.fileinfo.delete(v=vs.110).aspx) says that it can throw `IOException`, `SecurityException` and `UnauthorizedAccessException`, so I have removed `FileNotFoundException`, which cannot be thrown by it, added `IOException` to fix the bug and also added `SecurityException` just in case. [`FileInfo.Exists()`](https://msdn.microsoft.com/en-us/library/system.io.fileinfo.exists(v=vs.110).aspx) does not throw.

### DIFF
--- a/Core/Repositories/MachineCache.cs
+++ b/Core/Repositories/MachineCache.cs
@@ -104,7 +104,10 @@ namespace NuGet
                             packageFile.Delete();
                         }
                     }
-                    catch (FileNotFoundException)
+                    catch (IOException)
+                    {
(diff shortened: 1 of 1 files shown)


304NotModified:
Thanks!

PR #133, opened 2016-10-17
MonoAndroid必须带上版本号，否则在ProGet上发布后客户端找不到

tldzyx:


304NotModified:
We need more info if we like to merge this.

AFAIK `MonoAndroid` is the correct one, see 

https://github.com/dotnet/corefx/blob/master/Documentation/architecture/net-platform-standard.md

PR #87, opened 2016-04-14
Better handling of large packages

iHateInventNames:
Resolved OutOfMemory exceptions for multiple opening of packages >256Mb unpacked (tested on Wt)
Also, looks like it improves responsiveness while opening big packages.

PR #182, opened 2017-04-03
A quick fix for "Holds onto file handle on file close #145"

### LINKED ISSUE #145: Holds onto file handle on file close
- Bug
- 3.19
- ClickOnce

In case of a BUG:

- What is the current result?
Close file, still keeps the file handle open on the file. Jump into the command prompt and try to delete or move the file.

del : Cannot remove item C:\Development\Logging.1.0.0.nupkg: The process cannot access the file 'C:\Development\Logging.1.0.0.nupkg'
because it is being used by another process.

Close file closes the file handler.

- Are there any work arrounds?
Close NuGet Package Explorer

Not sure

Any

petterlabraaten:
A quick fix for "Holds onto file handle on file close #145"

### DIFF
--- a/Core/Packages/DataServicePackage.cs
+++ b/Core/Packages/DataServicePackage.cs
@@ -177,5 +177,9 @@ namespace NuGetPe
         {
             return this.GetFullName();
         }
+
+        public void Dispose()
+        {
+        }
     }
 }
--- a/Core/Packages/SimplePackage.cs
+++ b/Core/Packages/SimplePackage.cs
@@ -188,5 +188,9 @@ namespace NuGetPe
         }
 
         #endregion
+
+        public void Dispose()
+        {
+        }
     }
 }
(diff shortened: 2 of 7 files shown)


304NotModified:
This will fix the open stream on close of NPE isn't it?

petterlabraaten:
The question is for me? (I'm new to this). Yes. It fixes the problem where "File Handles" (Streams) are left unclosed when a local package is opened and then closed; preventing users from modifying/moving/deleting that packages without fully shutting down NPE.

304NotModified:
Thanks, will test it!

304NotModified:
Works great!

304NotModified:
Also tested with a large package (for https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/pull/87)  of 333MB. Works also well

304NotModified:
Fixes https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/issues/145

Issue #164, opened 2017-01-10
Default to opened package Location for 'Save As'


EddyCwdry:
**Type** : Feature Request

**NPE version**: Any

**Installed from**: ClickOnce

FEATURE REQUEST:
 
Currently when you open an existing package to update it, if you open/navigate to any other folder (eg using 'replace with'), then the 'Save As' prompt will then open in the last folder accessed, rather than the folder that the package was opened from.

Often I will open an old package, replace various files, then go save as, and currently I then have to navigate back to the base repo folder for our office. It's just a little convenience thing that would save quite a few clicks throughout my day; and to me, it seems to be what would be considered 'expected behavior' for a Save prompt.

304NotModified:
closed by #187 

thanks!

PR #187, opened 2017-04-12
Using the OpenSaveFileDialog will now default to the directory of the…

jjdotcc:
This should solve issue #164.

It's caused because the filename is included when setting the InitialDirectory property which as of Vista causes it to use a different path. [FileDialog.InitialDirectory - MSDN](https://msdn.microsoft.com/en-us/library/system.windows.forms.filedialog.initialdirectory(v=vs.110).aspx)


304NotModified:
Cool! Thanks!

Will test soon.

304NotModified:
nice enhancement! Thanks!

PR #194, opened 2017-04-29
WIP high contrast mode - improve accessibility

304NotModified:
don't merge

in this PR, normal mode:

image

image

image

image

image


wjk:
@304NotModified I have a question about this PR, and how the app will look once it is merged. I am quite fond of the default, VS2010-like styling of the application. While I understand the importance of a minimalist theme for accessibility purposes, I will never need to use it myself. In addition, I think that the default WPF themes are horrendously ugly, so I was very pleased to see alternative styling used in this application.

My bottom line is this: Will the current theme still be supported in the app once this PR is merged? [There is an API](https://docs.microsoft.com/en-us/dotnet/api/system.windows.systemparameters.highcontrast?view=netframework-4.7#System_Windows_SystemParameters_HighContrast) to detect whether or not high-contrast mode is enabled. Would it be possible to use this API to dynamically select which theme to use at app launch time? Thanks!

304NotModified:
Hi @wjk 

I have created this PR to see the changes, but I won't merge this one. This has to much impact on the current GUI and AFAIK the issue is solved with the fork.

Would be great to support high contrast in this repo, but I haven't found time to fix that (including the check). If you would like to help us with that, that would be great!

PR #212, opened 2017-07-20
#211 Sets parent first to force a path update.

### LINKED ISSUE #211: Some treeview items are duplicated
**Type** (choose one): Bug

**NPE version**:  3.23

**Installed from**:  chocolatey.org

When dragging certain folder names into a sub folder then back to the lib folder and then back to a sub folder it will duplicate the item.

image

jjdotcc:
Fixes #211 - Now updates the parent first to update the path so the compare works correctly. Also means items will be put back in alphabetical order.


304NotModified:
thanks!

PR #213, opened 2017-07-20
Changed PreviewMouseMove to MouseMove.

### LINKED ISSUE #207: Very hard to expand folders without moving folders around
**Type** (choose one):
- Bug

**NPE version**: (e.g. 3.12) (see Help -> about)
3.21.0.0

**Installed from**: ClickOnce / chocolatey.org / Zip on CodePlex / Windows Store
Not sure.

In case of a BUG:

Observed behavior:
Open a nuget package and try to expand folders, especially a very deep folder structure.  It is very hard to do this without accidentally moving a folder rather than just expanding the next level. The sensitivity of the drag and drop for folders is too small. Need to have it require more drag motion before it considers it a drag and drop. Once this happens you get errors displayed when you try to further expand levels.: "There is another file or folder with the same name as 'xxxx'".  Have to close app and reopen and start again.

It often takes me several minutes to successfully get  even 4 or 5 levels of folders expanded no matter how careful I try to be.

It seems like it may be much more pronounced issue when trying to expand the bottom folder in a tree. This is where I hit it most often. Just clicking the arrow to the left of the folder name easily drags and drops the folder to the grandparent folder.

Expected behavior:
Less sensitivity in detecting drag and drop. See Windows File Explorer for an example. That is the behavior people expect.

Workarounds:
Close app, reopen and try again.

jjdotcc:
Fixes #207. The expand arrows can no longer drag drop meaning it is easier to use the treeview.

### DIFF
--- a/PackageExplorer/PackageViewer.xaml
+++ b/PackageExplorer/PackageViewer.xaml
@@ -444,7 +444,7 @@
                         <EventSetter Event="DragOver" Handler="OnTreeViewItemDragOver" />
                         <EventSetter Event="Drop" Handler="OnTreeViewItemDrop" />
                         <EventSetter Event="PreviewMouseLeftButtonDown" Handler="PackagesTreeViewItem_PreviewMouseLeftButtonDown" />
-                        <EventSetter Event="PreviewMouseMove" Handler="PackagesTreeViewItem_PreviewMouseMove" />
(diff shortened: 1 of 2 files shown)


304NotModified:
Thanks! Will try to test soon!

Issue #229, opened 2017-09-28
No assembly info shown for .NET Standard 2.0 assemblies


luigiberrettini:
**Type** (choose one): bug

**NPE version**: 3.21.0.0

**Installed from**: ClickOnce

![assembly-info-in-package-explorer](https://user-images.githubusercontent.com/4513301/30962881-c3465d92-a44b-11e7-96f9-c75e578dedb1.png)

Assembly info taken from build DLL and DLL extracted from NuGet package (ILSpy CI build):
```csharp
[assembly: CompilationRelaxations(8)]
[assembly: RuntimeCompatibility]
[assembly: Debuggable()]
[assembly: TargetFramework(".NETStandard,Version=v2.0")]
[assembly: AssemblyCompany("Asbly company")]
[assembly: AssemblyConfiguration("Debug")]
[assembly: AssemblyCopyright("Asbly copyright")]
[assembly: AssemblyDescription("Asbly description")]
[assembly: AssemblyFileVersion("3.0.0.0")]
[assembly: AssemblyInformationalVersion("4.0.0.0")]
[assembly: AssemblyProduct("Asbly product")]
[assembly: AssemblyTitle("Asbly title")]
[assembly: NeutralResourcesLanguage("en")]
[assembly: AssemblyVersion("2.0.0.0")]
```

Related to [dotnet/sdk issue 1555](http://github.com/dotnet/sdk/issues/1555) and [NLog/NLog issue 1536](http://github.com/NLog/NLog/issues/1536)

304NotModified:
Thanks for submitting!

Reproducible with NLog 4.5 beta 2:


![image](https://user-images.githubusercontent.com/5808377/30963097-8a5b6b3e-a44c-11e7-8a3a-44b38f729f94.png)

304NotModified:
With NLog 5 beta 11

- netstandard 1.3 works
- netstandard 1.5 doesn't work

![image](https://user-images.githubusercontent.com/5808377/30963148-b42f8346-a44c-11e7-8b67-67da3ccda359.png)


![image](https://user-images.githubusercontent.com/5808377/30963155-bb798bec-a44c-11e7-9415-1b22f98d2bc2.png)

luigiberrettini:
Repo [dotnet-core-assembly-info](http://github.com/luigiberrettini/dotnet-core-assembly-info) available for investigation.

Try looking at ILSpy CI build code to understand how to solve this issue.

[...]

PR #235, opened 2017-10-24
Read assembly attributes for .NET Core and other runtime assemblies

zvirja:
Currently tool is unable to read assembly attributes if assembly targets some versions of the .NET Standard runtime. Consider the [AutoFixture 4.0 RC1](https://www.nuget.org/packages/AutoFixture/4.0.0-rc1) package as an example. If you select assembly targeting the .NET Standard, tool will not show the attributes.

I've adjusted the logic to perform more robust dependencies lookup. Also I use `Assembly.ReflectionOnlyLoadFrom(assemblyPath)` to load the assembly as this is the [advised by MS](https://docs.microsoft.com/en-us/dotnet/framework/reflection-and-codedom/how-to-load-assemblies-into-the-reflection-only-context) way.

Now it seems to work for all the packages I've tested.

@304NotModified @onovotny Please take a look and let me know if you see any issues with the PR.

### DIFF
--- a/CodeExecutor/AppDomainWorker.cs
+++ b/CodeExecutor/AppDomainWorker.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
+using System.Resources;
 
 namespace CodeExecutor
 {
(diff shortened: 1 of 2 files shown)


zvirja:
It's strange that build failed - I haven't touched anything related to the error.. 😕

onovotny:
I changed the base appveyor image to the release, can you push/rebase and see if that fixes it?

zvirja:
@onovotny Thanks 😉

> can you push/rebase and see if that fixes it?

Actually, you can re-run build by yourself without a need to re-push the commit - just open the [build report](https://ci.appveyor.com/project/NuGetPackageExplorer/nugetpackageexplorer/build/5.0.18%2bg213a2daa31) and press the `Re-Build PR` button  (you should be logged in as admin to AppVeyor):

image

Let me know if you don't have permissions so you still want me to re-push the commit.

onovotny:
I can't hit that button as I don't have access to that AppVeyor team.

zvirja:
@onovotny Thanks for clarifications 😉 I've rebased and re-pushed my code and now the build is green. However we know that in future version of VS it will get broken 😄

Now my PR is eligible for review, so @onovotny and @304NotModified please take a look.

304NotModified:
@onovotny shall I add you to the appveyor team?

Ps close/reopen PR also works ;)

304NotModified:
Fixes #299 isn't it?

zvirja:
> Fixes #229 isn't it?

I haven't searched for the issues before fixing this 😊Sorry for that.

Well, it appears to fix the issue partially. Everything is fine for .netstandard 1.5, but not for netstandard 2.0. It appears that in case of .NET Standard 2.0 it references the `netstandard 2.0` assembly which is not available 😢

onovotny:
@304NotModified please - my name there is the same as here, onovotny.

zvirja:
@304NotModified I've investigated that a bit more and found that the only reliable way to read those attributes would be to not use the System.Reflection. Instead, we should switch to something like [Mono.Cecil](https://www.nuget.org/packages/Mono.Cecil/0.10.0-beta6) which doesn't require dependencies for its work. That will be a long term investment as we will not become broken with .NET Standard 3.0 and so on.

Would you approve the PR if I redesign the approach to use `Mono.Cecil` instead of AppDomain and assembly load? If you wish, I could fire that in a separate PR, so you can evaluate how it looks.

304NotModified:
> Would you approve the PR if I redesign the approach to use Mono.Cecil instead of AppDomain and assembly load? If you wish, I could fire that in a separate PR, so you can evaluate how it looks.

Fine to me, what do you think @onovotny ?

304NotModified:
@onovotny you should have access AppVeyor now, I used the GitHub team (https://github.com/orgs/NuGetPackageExplorer/teams/admins/members), as I can't find how to add you a AppVeyor user

onovotny:
If Mono.Cecil is the right thing, then sure, go for it. Is that better than using `System.Reflection.Metadata`? That can load any assembly as well - I've used it in the past to get assembly information in a similar way: 
https://github.com/onovotny/ReferenceGenerator/tree/dev/src/ReferenceGenerator.Engine

@304NotModified I can now log into the AppVeyor team, but the group's permissions don't let me trigger a rebuild of a commit, just deploy things :)
image

onovotny:
The code here shows how to use `System.Reflection.Metadata` to read any assembly and show data about it:
https://github.com/onovotny/ReferenceGenerator/blob/dev/src/ReferenceGenerator.Engine/AssemblyInfoFile.cs

zvirja:
@onovotny Well, I briefly tried to use the `System.Reflection.Metadata` package and found that experience is not the best. While for the assembly name and reference it's not too hard, for the custom attributes it becomes complicated. The worse complication is that you need to implement your own `ICustomAttributeTypeProvider<>` to obtain the attribute values. While implementation might be [copied from here](https://github.com/dotnet/corefx/blob/bffef76f6af208e2042a2f27bc081ee908bb390b/src/System.Reflection.Metadata/tests/Metadata/Decoding/CustomAttributeDecoderTests.cs#L337), it's still 200 lines of code that we need to maintain. For the Cecil it's much easier and everything works out of the box.

However, I do like the `System.Reflection.Metadata` as it takes much less space - totally 1MB compared to 10MB for `Mono.Cecil`. I'll give it a try and will fire a separate PR.

onovotny:
Seems like it's worth a shot given the size difference, if it works.

zvirja:
@onovotny @304NotModified Take a look and #237. I've implemented the feature using the advised way. Looks pretty nice and I thought it would be more complicated 😉

If we decide to proceed on that direction, feel free to close this PR.

zvirja:
Closing this one in favor of #237.

PR #238, opened 2017-11-01
Include Packages folder in Types and Core

ridomin:
Adding Packages folder into csproj. While this issue https://github.com/dotnet/sdk/issues/1708 gets solved


onovotny:
The build is failing because it causes a duplicate items entry in 15.4, which is where the CI is run from. Can we condition the include on something?

onovotny:
Alternatively, update the image in the `AppVeyor.yml` to be `Visual Studio 2017 Preview` as I think 15.5 is there.

onovotny:
Included as part of #240, thanks.

PR #242, opened 2017-11-18
update version to 3.25

304NotModified:
@onovotny 

any idea if this update also the version in the about?

image

That was first in the commonAssemblyInfo.cs afaik


304NotModified:
PS: we need to update https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/blob/master/README-developers.md

onovotny:
It should be automatic based on the `version.json` file since that updates the assemblies and the dialog gets its version from there: https://github.com/onovotny/NuGetPackageExplorer/blob/ba06c34fbabef2fbe1cb63f03296fadb9332cebc/PackageExplorer/AboutWindow.xaml.cs#L34

304NotModified:
OK, thx!

onovotny:
was there a reason this wasn't merged?

304NotModified:
I didn't update the version.json, therefor I closed this one ;)

PR #245, opened 2017-11-20
Fix Semver versioning for Chocolatey packages

### LINKED ISSUE #243: Chocolatey version on AppVeyor is non-semver
It looks like the current versioning on AppVeyor for Chocolatey packages is non-semver.

image

e.g. it should be 3.25 or 3.25.0

304NotModified:
fixes #243

### DIFF
--- a/version.json
+++ b/version.json
@@ -1,5 +1,6 @@
 {
   "version": "3.24",
+  "buildNumberOffset": -1,
   "publicReleaseRefSpec": [
     "^refs/heads/master$", // we release out of master
     "^refs/heads/dev$", // we release out of develop


304NotModified:
@onovotny your opinion on this please, see also #243. 

I guess the version bump PR should be a without a merge commit to keep the "git height" 0

version.json
```diff
@@ -1,5 +1,6 @@
 {
   "version": "3.24",
+  "buildNumberOffset": -1,
```
onovotny: I wouldn't do this for the app. You can use `3.25.0` if that's what you really want for the release and then use `3.26.0-build.{height}` for subsequent builds.

304NotModified: OK, guess that's https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/pull/246 ?

PR #246, opened 2017-11-20
Fix patch version to 0, update to V3.25

### LINKED ISSUE #243: Chocolatey version on AppVeyor is non-semver

304NotModified:
fixes #243

### DIFF
--- a/version.json
+++ b/version.json
@@ -1,5 +1,5 @@
 {
-  "version": "3.24",
+  "version": "3.25.0",
   "publicReleaseRefSpec": [
     "^refs/heads/master$", // we release out of master
     "^refs/heads/dev$", // we release out of develop


304NotModified:
With the fixed 0, we could make a merge commit again - I guess?

onovotny:
Yes, but I would not keep that version there. That version should only be there for an actual release. Before/after the release, the height should be somewhere, so like `3.25.0-build.{height}`. Otherwise there's no way to tell builds apart.

onovotny:
Basically, the process for a release is:

1.  Change the version to a stable one like `3.25.0`.
2.  Build in CI, verify, etc. Release.
3.  Update version to `3.26.0.build.{height}`

304NotModified:
> Otherwise there's no way to tell builds apart.

I normally use git tags for that. Or [Fody.Stamp](https://github.com/Fody/Stamp)

onovotny:
The problem with git tags are that they can be modified. Tags can be deleted, modified, etc.

onovotny:
Alternatively, just let the third digit represent the build and you release when ready. So the released version is 3.25.2123. I don't see the issue with that, and if you look at version numbers for other products, like Visual Studio, it's similar. The build/patch are always incrementing, they're never "4.0.0" releases.

304NotModified:
> I don't see the issue with that

It's not semver and I find it confusing. 

> they're never "4.0.0" releases.

Well it's 2017.4 etc. 

> The problem with git tags are that they can be modified. Tags can be deleted, modified, etc.

true, but why do we need this proof? nuget.org is already proving that there is only one 3.25 version.

Maybe I'm missing something, but  3.25.2123 won't give any security gains.

onovotny:
2017.4 isn't the version. If anything it's VS 2017 15.4, and that's a marketing name. The actual version on the files is for 15.4.4 is `15.0.27004.2009`.

It's not about proof, it's about build reproducibility. I think there's generally value in having stable, reproducible builds. Checkout the same commit and it'll always give you the same build.

Having 3.25.2123 isn't about security, it's about reproducibility. If I checkout the commit that came from, I'll always get that version.

onovotny:
If you run the powershell, you'll see all of the values:

```
 $vers = & $env:UserProfile\.nuget\packages\nerdbank.gitversioning\2.0.41\tools\Get-Version.ps1
   Preview    update-nuget-client  PS> $vers


CloudBuildNumber                : 6.0.1+g150d8e2e7e
CloudBuildNumberEnabled         : True
BuildMetadataWithCommitId       : {g150d8e2e7e}
AssemblyVersion                 : 6.0.0.0
AssemblyFileVersion             : 6.0.1.3349
AssemblyInformationalVersion    : 6.0.1+g150d8e2e7e
PublicRelease                   : False
PrereleaseVersion               :
SimpleVersion                   : 6.0.1
BuildNumber                     : 1
MajorMinorVersion               : 6.0
GitCommitId                     : 150d8e2e7ee77c8ccb19112290df3ee7c01f3eeb
GitCommitIdShort                : 150d8e2e7e
VersionHeight                   : 1
VersionHeightOffset             : 0
Version                         : 6.0.1.3349
CloudBuildVersionVarsEnabled    : True
CloudBuildVersionVars           : {[GitAssemblyInformationalVersion, 6.0.1+g150d8e2e7e], [GitBuildVersion, 6.0.1.3349],
                                  [GitBuildVersionSimple, 6.0.1]}
BuildMetadata                   : {}
BuildMetadataFragment           : +g150d8e2e7e
NuGetPackageVersion             : 6.0.1-g150d8e2e7e
NpmPackageVersion               : 6.0.1-g150d8e2e7e
SemVer1                         : 6.0.1-g150d8e2e7e
SemVer2                         : 6.0.1-g150d8e2e7e
SemVer1NumericIdentifierPadding : 4

```

You can see that some things have three parts, others have four, etc. Others, like the AssemblyVersion, by default, stay at major minor but with four parts.

304NotModified:
> 2017.4 isn't the version. If anything it's VS 2017 15.4, and that's a marketing name. The actual version on the files is for 15.4.4 is 15.0.27004.2009.

Good one 2017 15.4 indeed.

> You can see that some things have three parts, others have four, etc. Others, like the AssemblyVersion, by default, stay at major minor but with four parts.

I agree that 3-4 parts are nice, but it should not be in the package version (on nuget or chocolatey) IMO.

304NotModified:
anyway, 

if we set ` "buildNumberOffset": -1,`, then we have the best of both worlds? (semver, reproducable builds, clear versioning)

onovotny:
That's fine. I don't really want to get hung up on the details :) Versions to me are about conveying intent.

304NotModified:
superseded by  #254

PR #240, opened 2017-11-16
Update nuget client

onovotny:
Branch to move to the newer Nuget.Client. This builds but needs a thorough review of all editors for type conversion. Also needs review of all XAML bindings due to internal OM changes.

Fixes #224 & #230 

This will be the first of several PR's implementing features based on the new client libraries. Initially, this PR is just to preserve existing functionality using the new libraries.

PR #252, opened 2017-11-24
Move to NuGet.Packaging for creation and reading of packages

### LINKED ISSUE #251: Use Packaging Object Model for Save / NuSpec
#240 updated the code to leverage the Packaging OM for opening `nupkg` files. This is to track the remainder of the work.

Success criteria for this is that direct XML serialization is removed.

onovotny:
Builds on #240 and also fixes #251 

With this PR, NPE is fully using the NuGet.Packaging types for manifest reading/saving.


onovotny:
@304NotModified This PR should supersede #240.

I believe it's ready to go, though I'm sure there'll be bugs along the way to fix. The smoke test appears to work. I can edit/save/open packages, etc.

PR #254, opened 2017-11-27
CI Builds

### LINKED ISSUE #253: Add "Nightly" release channel
Goal is to have an `appx` that can be sideloaded on machines as part of the CI build. This will help let people use it before we're ready for a "stable" release to the store.

Requirements:
- Must be SxS with the store version

Due to manifest requirements, we'll need a new packaging project with a new manifest that uses a different package id. An overlay on the icon to tell the difference would be nice.

Also for versioning, we'll be incrementing the patch for every build since the packages can't use "prerelease" labels. That means, builds will be like `6.0.48`, etc. Once deemed stable, the released version will go to the store.

The CI process will ensure the bits are signed with a certificate to a publically trusted root CA, so no issues with sideloading.

/cc @304NotModified

onovotny:
This requires #252 to be merged first.

This PR adds a CI appxbundle and badges. It's also generating choco packages, but those aren't yet externally published.

Fixes #253


304NotModified:
please rebase on master, by skipping the 40 commits. Thx :)

PR #256, opened 2017-11-29
Added an option to the publish window to use Credentials.

ironicnet:
Added an option to the publish window so we can choose to send the request with the default credentials or setting the Personal Access Token in the credential.

I think that this fixes #58 . Although I think that the sources should have a way to identify which kind of authentication to use.

I don't know if i broke any coding / design guideline as I didn't see any guide for contributing. I just followed some code.
Also, didn't find any unit test or test project to add tests to it.
Tested it against our VSTS and worked well.
Preview of the window:
image

PR #264, opened 2017-12-07
Token workaround

### LINKED ISSUE #225: Package versions are listed in alphabetical order instead of version order
**Type** (choose one): Bug

**NPE version**: 3.23

- What is the current result?

When I click "show all versions" in the "Select package" dialog, the versions of a package are listed in descending alphabetical order, instead of descending version order, e.g. 1.10 is considered lower than 1.2.

Versions should be listed in descending version order

- Please post full exception details in case of an Exception (message, stacktrace, inner exceptions)

N/A

- Are there any work arrounds? yes/no

Not as far as I can tell

I don't know

Any package with version component (major minor or patch) great than or equal to 10. Newtonsoft.Json shows the problem for instance; when you click "show all versions", you see 9.x, then 8.x, etc, and 10.x is at the end of the list.

### LINKED ISSUE #255: Add support for $version$ with new object model
#252 breaks current support for the $version$ token since code is read through the NuGet object model.

https://github.com/NuGet/Home/issues/6228 tracks this on the NuGet side, but we'll attempt to workaround it a different way here.

onovotny:
Fixes #255 

Basically, this works if a token represents a full version string, but I think it’s possible to use a token anywhere in the string, like 3.3.0$versionsuffix$. This won't work in that case, but I don't think that's too common.

If there are any tokens in the versions, save package and publish commands are disabled since that's not a valid package. It can only be saved as a manifest.

Here's a nuspec I used for testing

```xml
<?xml version="1.0" encoding="utf-8"?>
<package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
  <metadata>
    <id>MyPackage</id>
    <version>$version$</version>
    <title></title>
    <authors>oren</authors>
    <owners>oren</owners>
    <requireLicenseAcceptance>false</requireLicenseAcceptance>
    <description>My package description.</description>
    <dependencies>
      <group>
        <dependency id="foo" version="$versio1n$" />
        <dependency id="bar" version="$version$" />
        <dependency id="bb" version="[$version$, $version4$)" />
      </group>
      <group targetFramework=".NETFramework4.5">
        <dependency id="bb" version="4.5.0" />
      </group>
    </dependencies>
  </metadata>
  <files>
  </files>
</package>
```

### DIFF
--- a/Core/Utility/ManifestUtility.cs
+++ b/Core/Utility/ManifestUtility.cs
@@ -12,8 +12,8 @@ namespace NuGetPe.Utility
 {
     public static class ManifestUtility
     {
-        public const string TokenMetadataStart = "0.0.0-TOKENSTART.";
-        public const string TokenMetadataEnd = ".TOKENEND";
+        const string TokenMetadataStart = "0.0.0-TOKENSTART.";
+        const string TokenMetadataEnd = ".TOKENEND";
         static readonly Regex tokenRegex = new Regex(@"([$])(?:(?=(\\?))\2.)*?\1", RegexOptions.Compiled);
         static readonly Regex metadataRegEx = new Regex(@"0\.0\.0\-TOKENSTART\.([^.]+)\.TOKENEND", RegexOptions.Compiled);
 
--- a/Core/Utility/ManifestUtility.cs
+++ b/Core/Utility/ManifestUtility.cs
@@ -11,10 +11,10 @@ namespace NuGetPe.Utility
 {
     public static class ManifestUtility
     {
-        public const string TokenMetadataStart = "0.0.0+TOKENSTART.";
+        public const string TokenMetadataStart = "0.0.0-TOKENSTART.";
         public const string TokenMetadataEnd = ".TOKENEND";
         static readonly Regex tokenRegex = new Regex(@"([$])(?:(?=(\\?))\2.)*?\1", RegexOptions.Compiled);
-        static readonly Regex metadataRegEx = new Regex(@"0\.0\.0\+TOKENSTART\.([^.]+)\.TOKENEND", RegexOptions.Compiled);
+        static readonly Regex metadataRegEx = new Regex(@"0\.0\.0\-TOKENSTART\.([^.]+)\.TOKENEND", RegexOptions.Compiled);
 
         public static Stream ReadManifest(string file)
         {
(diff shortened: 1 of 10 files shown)


Core/Utility/ManifestUtility.cs
```diff
@@ -0,0 +1,136 @@
...
+using NuGet.Versioning;
+
+namespace NuGetPe.Utility
+{
+    public static class ManifestUtility
+    {
+        const string TokenMetadataStart = "0.0.0-TOKENSTART.";
+        const string TokenMetadataEnd = ".TOKENEND";
+        static readonly Regex tokenRegex = new Regex(@"([$])(?:(?=(\\?))\2.)*?\1", RegexOptions.Compiled);
+        static readonly Regex metadataRegEx = new Regex(@"0\.0\.0\-TOKENSTART\.([^.]+)\.TOKENEND", RegexOptions.Compiled);
```
304NotModified: do we really need a regex? I prefer `IndexOf` for simplicity and more clear performance.

Core/Utility/ManifestUtility.cs
```diff
@@ -0,0 +1,136 @@
...
+        {
+            return ReadManifest(File.OpenRead(file));
+        }
+
+
+        public static bool IsTokenized(this NuGetVersion version)
+        {
+            var labels = version.ReleaseLabels.ToList();
+
+            return labels.Count >= 3 && labels[0] == "TOKENSTART" && labels[labels.Count - 1] == "TOKENEND";
```
304NotModified: Would be nice to create consts for this

Core/Utility/ManifestUtility.cs
```diff
@@ -0,0 +1,136 @@
...
+
+            // Get dependency nodes
+            var deps = xdoc.Root.Descendants(ns + "dependency");
+            foreach (var dep in deps)
+            {
+                var val = dep.GetOptionalAttributeValue("version");
+                if (!string.IsNullOrWhiteSpace(val))
+                {
+                    dep.SetAttributeValue("version", ReplaceMetadataWithToken(val));
+                }
```
304NotModified: could we deduplicate this?

onovotny: deduplicate what?

Core/Utility/ManifestUtility.cs
onovotny: The RegEx was the easiest to get the entire token span, start to end, also accounting for multiple token values in a string (which is valid for the VersionRange on a dependency group).

I didn't see any perf issues here as it's also a UI app, not a server handling tons of requests :)

onovotny:
I'm not sure a legacy project like this is worth adding a suite of unit tests around...

304NotModified:
will test this after #256

onovotny:
Merging this in for now; will address any subsequent issues as they arise.

PR #284, opened 2017-12-16
Restore Nuget packages before building.

KirillOsenkov:


onovotny:
I think the build.cmd and build.proj can just be nuked? They're not really used by anything anymore.

Issue #248, opened 2017-11-22
Support Package Signing


ridomin:
**Type**:  Feature request
**NPE version**: 3.21.0.0
**Installed from**:  Windows Store

NPE should be updated to be aligned with NuGet Package signing https://github.com/NuGet/Home/issues/2577 

I'd suggest the next User Stories by priority

- P0. Disable edits if the package is signed.
- P1. Show signature information.
- P1. Allow removing the signature when saving.
- P2. Allow signing packages.

maartenba:
Is there any news on signing into for non-MS projects?

ridomin:
Hi @maartenba 
We are finishing the signature format spec, although we have started the implementation in the https://github.com/NuGet/NuGet.Client/commits/dev-feature-signing branch

We will announce progress once we publish the spec and have beta bits with signing support.

maartenba:
Thanks!

onovotny:
@maartenba I have a client app impl that uses Key Vault to do the sign operations and uses Bouncy Castle to generate the SignedCms message (since the APIs in the .NET one to do what I need don't exist yet):
https://github.com/onovotny/NuGetKeyVaultSignTool

onovotny:
P0 is done, P1 is partial (show signature information shows the publisher, but not validation status yet).

onovotny:
We now support showing signature information. Removing signature info and signing will likely wait until the next release.

onovotny:
The P0 and P1 items are both done now.

onovotny:
Closing out this issue as #292 tracks signing packages from NPE and the rest of the items here have been completed.

PR #290, opened 2017-12-17
For signed packages, don't resave during publish

onovotny:


onovotny:
Also adds signature removal for #248

PR #328, opened 2018-01-27
unescape zip file names

### LINKED ISSUE #327: Package entry names not properly decoded
**Type**: Bug

**NPE version**: Latest

**Installed from**: Compiled from source!

In case of a BUG:

- What is the current result?
If a package contains files or folders with special characters they are not properly decoded.
Example: `portable-net40+sl5+win8+wp8+wpa81` is displayed as  `portable-net40%2Bsl5%2Bwin8%2Bwp8%2Bwpa81`

It should be properly decoded: `portable-net40+sl5+win8+wp8+wpa81`

The bug was introduced by https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/pull/272
The [`PackageArchiveReader` did decode the zip entry names](https://github.com/NuGet/NuGet.Client/blob/91023394890b1458ec0c5940128da73e04089869/src/NuGet.Core/NuGet.Packaging/PackageExtraction/ZipArchiveExtensions.cs#L31-L45) which now does not happen.

image

campersau:
this was previously done by the PackageArchiveReader

fixes #327

### DIFF
--- a/Core/Packages/ZipPackageFile.cs
+++ b/Core/Packages/ZipPackageFile.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Diagnostics;
 using System.IO;
 using System.IO.Compression;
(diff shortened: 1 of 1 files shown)


onovotny:
thanks!

PR #326, opened 2018-01-27
support NuGet v3 feeds

campersau:
By using the official NuGet Client API for package searching / publishing.

- Use `https://api.nuget.org/v3/index.json` feed by default
- [Credentials support](https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/pull/326#issuecomment-361408998) for every operation (search / download / publishing)
- Recent files updated to work with v3 feeds
- Download progress enhancements. Before there were to many UI updates which blocked the progress bar from properly updating.
- "Append V2 Api To Url" in publish dialog is no longer necessary

This has some drawbacks:
- no sorting
- no package size (see https://github.com/NuGet/NuGetGallery/issues/2957)
- no total package count
- no package hash (it is possible with [`IPackageDownloader`](https://github.com/NuGet/NuGet.Client/blob/8144fdeca466151c1516b65dc30d429562890b86/src/NuGet.Core/NuGet.Packaging/Definitions/IPackageDownloader.cs#L59) but it requires to download the nupkg which makes it useless)
- no version download count ([see here](https://github.com/NuGet/NuGet.Client/blob/998012e9c32478c2709b6b132c8b6a11a30166f9/src/NuGet.Core/NuGet.Protocol/Resources/PackageSearchResourceV3.cs#L43-L54))
- publish date only in "show all versions" for v3 feeds
- ~~no download progress (maybe https://github.com/NuGet/Home/issues/5095 will work)~~ it worked!

~~This is a WIP because I haven't checked if authentication actually works, also I would like to get some feedback on this.~~

fixes #50

PR #332, opened 2018-01-28
remove BindingGroup and use property bindings for validation instead

### LINKED ISSUE #331: Publish Package does not work
**Type**: Bug

**NPE version**: master

**Installed from**: Compiled from source!

- What is the current result?
Publishing a package does not work, because the publish button is always disabled.

Publish button should not be disabled if all required data is available.

The bug was introduced by https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/pull/256.
The publish dialog is using a `BindingGroup` for validation so bindings are not applied until [`UpdateSources`](https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/blob/a75b6d687f40e520085b3e80df8520cb5437a7a9/PackageExplorer/PublishPackageWindow.xaml.cs#L20) is called. So the [`IsAuthSet`](https://github.com/ironicnet/NuGetPackageExplorer/blob/a9d485c4fe0c11fecb11118d158b0ee1cb5ab287/PackageExplorer/PublishPackageWindow.xaml#L116) property does never change. 

image

campersau:
fixes #331


onovotny:
Closing since this is handled by #326

PR #340, opened 2018-02-13
show download count per version

campersau:
It is quite hacky because it only works for V3 feeds.

Using the suggested implementation from https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/pull/326#issuecomment-365362992

image


onovotny:
thanks!

PR #341, opened 2018-02-14
Ensure dialog is always disposed

onovotny:


campersau:
When is it not null? It should always be closed and nulled:

https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/blob/a0b6d312212f027e6d428c922ba47b1a0fa7580a/PackageExplorer/MefServices/PackageDownloader.cs#L152-L164

onovotny:
You would think... but I see some crash reporting in the store dev center that points to the finalizer getting called:

```

Frame | Image | Function | Offset
-- | -- | -- | --
0 | mscorlib.ni.dll | System.StubHelpers.StubHelpers.GetCOMIPFromRCW | 0x0000000000000001
1 | Ookii.Dialogs.Wpf.dll | Ookii.Dialogs.Wpf.Interop.IProgressDialog.StopProgressDialog | 0x0000000000000001
2 | Ookii.Dialogs.Wpf.dll | Ookii.Dialogs.Wpf.ProgressDialog.Dispose | 0x000000000000003A
3 | Ookii.Dialogs.Wpf.dll | Ookii.Dialogs.Wpf.ProgressDialog.Finalize | 0x0000000000000010

```

PR #345, opened 2018-02-14
hide download button for local packages

campersau:
This was a regression after the V3 update.


304NotModified:
thanks for all the work @campersau !

campersau:
We can also keep the "download" link for local files because it is supported by NuGet.Client.
It would just do a file copy of the `*.nupkg` to the new location.

onovotny:
Thanks!

Issue #348, opened 2018-02-16
ID search shows incorrect results on NuGet


svick:
**Type**: Bug

**NPE version**: 4.0.12

**Installed from**: Windows Store

- What is the current result?

    When I search on NuGet using `id`, the search results seem to show an entry for each version of the package. This is especially problematic because it shows the oldest versions, not the newest. For example, searching for `id:Microsoft.CodeAnalysis`:

    ![](https://user-images.githubusercontent.com/287848/36322436-852361b0-134d-11e8-9149-39defd94dbd4.png)

- What is the expected result?

    Show a single entry for the package, just like it currently happens on MyGet:

    ![](https://user-images.githubusercontent.com/287848/36322500-b7dcee82-134d-11e8-9ee5-d91d645d1825.png)

- Are there any workarounds?

  If the goal is to get to the most recent version, selecting one of the entries and choosing "show all versions" works.

- Is there a version in which it did worked?

    I think this did work in the past, but I haven't checked older versions. Though I suspect that since MyGet works fine, this is probably a bug in nuget.org, not in NPE.

onovotny:
The latest Nightly version shows me this:

![image](https://user-images.githubusercontent.com/1427284/36347827-b31a9556-142f-11e8-9eb8-89193c8f4cf9.png)

onovotny:
Can you please try this again with the latest nightly? Using the v3 feed for nuget.org, I see this

![image](https://user-images.githubusercontent.com/1427284/36347842-3764a0f4-1430-11e8-952f-7700f59786aa.png)

onovotny:
We're just passing that string to the NuGet search API. It may have changed for the v3 endpoint. @joelverhagen, can you shed some light on this?

joelverhagen:
`packageid:` is exact match. `id:` is partial match. This is a query rule on NuGet.org, not sure how MyGet and VSTS observe these parameters. Note that the [official search API](https://docs.microsoft.com/en-us/nuget/api/search-query-service-resource#request-parameters) describes the supported query parameters, like `q`, `skip` and `take`, but does not mandate how the query term (`q`) is parsed. This could be improved in the future but so far it's not part of the specification today.

Here are some docs for other search terms:
https://docs.microsoft.com/en-us/nuget/consume-packages/finding-and-choosing-packages#search-syntax

onovotny:
[...]

PR #351, opened 2018-02-18
use packageId for exact matches

campersau:
### NuGet V2
image

### NuGet V3
image

---

### MyGet V2
image

### MyGet V3
image

---

### VSTS V3
image
Does not work with neither `packageId` or `id`.

### VSTS V2
image
Does not work with neither `packageId` or `id`.

---

### Local Feed
Does not work with neither `packageId` or `id`.

---

As suggested here https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/issues/348#issuecomment-366490686

fixes https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/issues/348


onovotny:
thanks!

PR #353, opened 2018-02-18
accept publish key or PAT in PublishDialog

campersau:
Publishing packages requires either a `Publish Key` or a `PAT` (Personal Access Token).
The implementation is not really nice. Maybe someone else has a good idea?

image

The UI currently always shows the abbrevation (PAT). Should the first occurrence be changed to `PAT (Personal Access Token)` ?

### DIFF
--- a/PackageExplorer/PublishPackageWindow.xaml
+++ b/PackageExplorer/PublishPackageWindow.xaml
@@ -28,10 +28,10 @@
         </DockPanel.BindingGroup>
 
         <!-- Main instruction -->
-        <TextBlock DockPanel.Dock="Top" Text="Enter your publish key." FontSize="12pt" Margin="10,10,0,15" Foreground="{StaticResource TaskDialogMainInstructionBrush}" />
+        <TextBlock DockPanel.Dock="Top" Text="Enter your publish key or PAT." FontSize="12pt" Margin="10,10,0,15" Foreground="{StaticResource TaskDialogMainInstructionBrush}" />
 
         <!-- Secondary instruction -->
-        <TextBlock DockPanel.Dock="Top" Margin="10,0" Text="Provide the publish Url and the publish key associated with it." TextWrapping="Wrap" />
+        <TextBlock DockPanel.Dock="Top" Margin="10,0" Text="Provide the publish Url and the publish key or PAT associated with it." TextWrapping="Wrap" />
 
         <!-- Id & Version -->
         <TextBlock DockPanel.Dock="Top" Margin="10,10,10,0">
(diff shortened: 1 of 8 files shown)


PackageViewModel/CredentialManagerProvider.cs
```diff
@@ -23,16 +26,16 @@ public Task<CredentialResponse> GetAsync(Uri uri, IWebProxy proxy, CredentialReq
...
 
-            var credentials = _credentialManager.Value.Get(uri);
+            var credentials = _credentialManager.Get(uri);
 
             if (credentials != null)
             {
                 return Task.FromResult(new CredentialResponse(credentials));
             }
-            return Task.FromResult(new CredentialResponse(CredentialStatus.ProviderNotApplicable));
+            return Task.FromResult(new CredentialResponse(CredentialStatus.UserCanceled));
```
campersau: Using `UserCanceled` here to allow the CredentialProvider to be called multiple times. Even if this time there are no credentials. When using `ProviderNotApplicable` then the CredentialProvider is never used again for this feed.

PackageViewModel/PackageViewModel.cs
```diff
@@ -28,6 +28,7 @@ public sealed class PackageViewModel : ViewModelBase, IDisposable
         private readonly IList<Lazy<IPackageRule>> _packageRules;
         private readonly ISettingsManager _settingsManager;
         private readonly IUIServices _uiServices;
+        private readonly CredentialPublishProvider _credentialPublishProvider;
```
campersau: Should this be an interface?

```c#
public interface ICredentialPublishProvider
{
    string PersonalAccessToken { get; set; }
}
```

PackageExplorer/PublishPackageWindow.xaml
```diff
@@ -59,8 +59,8 @@
 
         <!-- Publish key -->
         <DockPanel x:Name="PanelApiKey" DockPanel.Dock="Top" Margin="7,0,10,5">
-            <Label DockPanel.Dock="Left" Target="{Binding ElementName=PublishKey}" Content="Publish _key:" Width="75" />
-            <TextBox x:Name="PublishKey" VerticalContentAlignment="Center" IsEnabled="{Binding CanPublish}" Margin="0" Text="{Binding PublishKey}">
+            <Label DockPanel.Dock="Left" Target="{Binding ElementName=PublishKey}" Content="Publish _key or PAT:" Width="110" />
```
onovotny: Using Personal Access Token here does seem more clear

PackageViewModel/PackageViewModel.cs
onovotny: What other scenarios do you see where the interface makes sense.

PackageExplorer/PublishPackageWindow.xaml
campersau: The problem is that it is quite long:
image

What about this?
image

PackageViewModel/PackageViewModel.cs
campersau: Currently no one 😄

PackageExplorer/PublishPackageWindow.xaml
onovotny: that looks good, just the first time.

PackageViewModel/PackageViewModel.cs
onovotny: I'd say let's leave it as-is for now and we can always extract an interface if there's a reason to

PackageExplorer/PublishPackageWindow.xaml
campersau: Done.

onovotny:
Thanks!

Issue #358, opened 2018-02-23
Cannot build out of the box


matyasforian:
When I clone and try to build the project (either in Visual Studio or via the build.cmd) the build fails with errors that the class PackageExplorer.Resources is missing. 

To solve it I had to right click Resources.resx, click Properties and here set Custom Tool namespace to  "PackageExplorer.Resources" 

I guess you forgot to check this setting in to the .csproj file?

onovotny:
All of the files are checked in. What version of VS are you using? Do you have automatic package restore enabled? The resx files get set by targets that come down during restore.

svick:
As far as I can tell, this was caused by https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/commit/6531156bbb8f2cfeb6331db4e09d161020e33506. I have just created https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/pull/360, which should fix this.

matyasforian:
Yep, setting CustomToolNamespace in the csproj file fixes it.

PR #360, opened 2018-02-23
Fixed namespace for resources

svick:
Fixes https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/issues/358.

<s>Effectively undoes https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/commit/6531156bbb8f2cfeb6331db4e09d161020e33506.

This code was created by VS when I set the namespace in the properties of the Resources.resx file. It's possible the changes could be shorter (maybe just `<ItemGroup><EmbeddedResource Update="Resources.resx"><CustomToolNamespace>PackageExplorer.Resources</CustomToolNamespace></EmbeddedResource></ItemGroup>`), but I don't know if VS would like that, so I haven't even tried.</s>


onovotny:
@svick Update would work, it's "just" metadata. Just trying to think about it, wondering if there's a way to default it with a calculation instead. That logic should go here: https://github.com/onovotny/MSBuildSdkExtras/blob/master/MSBuild.Sdk.Extras/build/MSBuild.Sdk.Extras.Common.targets#L21-L28

svick:
@onovotny Maybe something like the following?

```xml
CustomToolNamespace="$(RootNamespace).Resources"
```

This assumes NPE has `RootNamespace` set to `PackageExplorer`, which it currently doesn't.

(Keep in mind I have very little experience with MSBuild.)

onovotny:
@svick yeah, I"m thinking something along the lines of getting the relative path and replacing the `\` with a `.`, which is effectively what VS does.

svick:
@onovotny What's wrong with using `RootNamespace`? It's what the resx tool uses by default and it should work fine for most projects. The reason it doesn't currently work for NPE is because the csproj name and the namespace aren't the same and because the code expects resources to be in the `.Resources` sub-namespace.

onovotny:
@svick nothing wrong with it, I just seem to recall it being root + folder name? I could be not remembering correctly either and just need more ☕

svick:
@onovotny What if I set `RootNamespace` to `PackageExplorer` and update all files to use `PackageExplorer.Resources` (instead of `PackageExplorer.Resources.Resources`)? That way, no changes to MSBuildSdkExtras are necessary and this project will use the conventional naming convention. I have updated this PR with that.

Though every time I select "Run Custom Tool", VS puts both `ItemGroup`s back into the csproj.

onovotny:
thanks!

Issue #292, opened 2017-12-18
Allow signing packages


onovotny:
Continuation from #248, NPE should support signing packages.

Need to think about:
- How to choose key
  - Always use Chooser UI?
  - PFX files with password?
- Always prompt? Way to save preferences (per package?)

PR #371, opened 2018-02-25
add sign package dialog

campersau:
Add support for signing packages using a sign package dialog.
Added new `Sign and Save As...` menu item.

- ~~Currently does not update the currently editing package metadata.~~

- ~~Currently always using the default hash algorith name: [SHA256](https://github.com/NuGet/NuGet.Client/blob/2847a3808d3cbcdb27fdfb9dfbaf95cd3ed95a57/src/NuGet.Clients/NuGet.CommandLine/Common/CommandLineUtility.cs#L55)~~

- ~~Currently no option to provide a custom [ITimestampProvider](https://github.com/NuGet/NuGet.Client/blob/dde7f1939f5c212f7db9b8e8576cf10414f6c1c3/src/NuGet.Core/NuGet.Packaging/Signing/Timestamp/Rfc3161TimestampProvider.cs)~~

- ~~Currently the last choosen certificate is not remembered during a NuGetPackageExplorer session.~~

- ~~Currently does not remember the last used certificate across NuGetpackageExplorer sessions.~~

Copied a lot from https://github.com/NuGet/NuGet.Client/blob/4c0c9658445573845ddbeff5656e4b3129f727a2/src/NuGet.Core/NuGet.Commands/SignCommand/SignCommandRunner.cs

image

image

image

image

fixes https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/issues/292


onovotny:
We should also support picking a certificate from the local certificate store. That search should filter by EKU's to only show code signing certs.

onovotny:
I'm not sure we need a custom TimestampProvider impl, but we should require a timestamp server to be provided and not allow packages to be created without a timestamp as that's just always a bad idea. The NuGet code has impls for that.

PackageExplorer/MefServices/UIServices.cs
```diff
@@ -204,6 +204,24 @@ public bool OpenSignatureValidationDialog(object viewModel)
...
+        {
+            var dialog = new SignPackageDialog
+            {
+                Owner = Window.Value,
+                DataContext = viewModel
+            };
+
+            if (viewModel is IDisposable disposable)
+            {
+                dialog.Closed += OnDialogClosed;
```
onovotny: Don't forget to unsubscribe from this event in the closed method

PackageExplorer/SignPackageDialog.xaml.cs
```diff
@@ -0,0 +1,38 @@
...
+        public SignPackageDialog()
+        {
+            InitializeComponent();
+        }
+
+        public string SignedPackagePath { get; private set; }
+
+        private void OnCertificatePasswordChange(object sender, EventArgs args)
+        {
+            var viewModel = (SignPackageViewModel)DataContext;
```
onovotny: Would it be better to just have a strongly-typed property on the dialog that does this? `ViewModel => (SignPackageViewModel)DataContext`?

PackageViewModel/SignPackageViewModel.cs
```diff
@@ -0,0 +1,363 @@
...
+                    {
+                        if (CertificateUtility.IsValidForPurposeFast(certificate, Oids.CodeSigningEku))
+                        {
+                            collection.Add(certificate);
+                        }
+                    }
+
+                    var certificates = X509Certificate2UI.SelectFromCollection(
+                        collection,
+                        "Choose a Certificate for Package Signing",
```
onovotny: These strings (and others on the dialogs as well) should be put in resources so they can be localized

PackageViewModel/SignPackageViewModel.cs
```diff
@@ -0,0 +1,363 @@
...
+                using (var signRequest = new SignPackageRequest(tempCertificate, GetHashAlgorithmName(tempCertificate)))
+                {
+                    SigningUtility.Verify(signRequest, NullLogger.Instance);
+
+                    var packagePath = _packageViewModel.GetCurrentPackageTempFile();
+                    var originalPackageCopyPath = Path.GetTempFileName();
+
+                    File.Copy(packagePath, originalPackageCopyPath, overwrite: true);
+
+                    using (var packageReadStream = File.OpenRead(packagePath))
```
onovotny: The code in here should support overwriting signatures; there's a slightly different code path for that if a package is already signed.

PackageViewModel/SignPackageViewModel.cs
```diff
@@ -0,0 +1,363 @@
...
+                    using (var packageWriteStream = File.Open(originalPackageCopyPath, FileMode.Open))
+                    using (var package = new SignedPackageArchive(packageReadStream, packageWriteStream))
+                    {
+                        var signer = new Signer(package, new X509SignatureProvider(null));
+                        await Task.Run(() => signer.SignAsync(signRequest, NullLogger.Instance, CancellationToken.None));
```
onovotny: Why do we need a Task.Run here?

PackageViewModel/SignPackageViewModel.cs
```diff
@@ -0,0 +1,363 @@
...
+                return null;
+            }
+            finally
+            {
+                ShowProgress = false;
+                CanSign = true;
+            }
+        }
+
+        private void ValidateCertificate()
```
onovotny: I'm not completely following why this needs a timer with a semaphore?

PackageViewModel/SignPackageViewModel.cs
```diff
@@ -0,0 +1,363 @@
...
+                OnError(ex);
+            }
+            finally
+            {
+                _certificateValidationSemaphore.Release();
+            }
+        }
+
+        // https://github.com/NuGet/NuGet.Client/blob/894388a598834a1fe8a0e483a2bc050c05693313/src/NuGet.Core/NuGet.Packaging/Signing/Utility/CertificateUtility.cs#L112-L124
+        private static HashAlgorithmName GetHashAlgorithmName(X509Certificate2 certificate)
```
onovotny: The hash algorithm chosen shouldn't have anything to do with the certificate. We should either hard-code Sha256 or put a drop-down dialog with the options and default to Sha256.

PackageViewModel/SignPackageViewModel.cs
```diff
@@ -0,0 +1,363 @@
...
+using System;
+using System.IO;
+using System.Security.Cryptography.X509Certificates;
+using System.Threading;
+using System.Threading.Tasks;
+using System.Windows.Input;
+
+namespace PackageExplorerViewModel
+{
+    public class SignPackageViewModel : ViewModelBase, IDisposable
```
onovotny: The ViewModel should have a property for the Timestamp Server. That can be stored in the application settings for reuse.

PackageExplorer/MefServices/UIServices.cs
campersau: See https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/pull/371/files/b2a065e26fecab7ffb260d6c0bb5c2ec6e7efa3d#diff-50954ff97bc9dd421a2b5f53498b8c05R232

PackageExplorer/SignPackageDialog.xaml.cs
campersau: I can do that. I just copied it from https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/blob/cdf3abacfc66363b49f6ccbb1968334aa8eb1948/PackageExplorer/PublishPackageWindow.xaml.cs#L23

PackageViewModel/SignPackageViewModel.cs
campersau: Will do. There a currently a lot of static strings in other dialogs as well.

PackageViewModel/SignPackageViewModel.cs
campersau: Yes, I haven't tested it yet.

PackageViewModel/SignPackageViewModel.cs
campersau: Otherwise it blocks the UI. I currently don't know why. Looks like it is the same problem as here https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/blob/0fb9b0869bf9f1136000700b87414f687728525b/PackageViewModel/EditablePackageMetadata.cs#L75

PackageViewModel/SignPackageViewModel.cs
campersau: Currently this is run on every keypress in the password field.
And the validation might take some time which resulted in an invalid state.

PackageViewModel/SignPackageViewModel.cs
campersau: Will do.

PackageExplorer/MefServices/UIServices.cs
onovotny: Sorry, didn't see that when looking online, GitHub sometimes hides too much :)

PackageExplorer/MefServices/SettingsManager.cs
```diff
@@ -132,6 +132,12 @@ public string SigningCertificate
             set => Settings.Default.SigningCertificate = value;
         }
 
+        public string TimestamperServer
```
onovotny: Nit: lets call it `TimestampServer`

campersau:
Timestamp Server (using [Rfc3161](https://github.com/NuGet/NuGet.Client/blob/894388a598834a1fe8a0e483a2bc050c05693313/src/NuGet.Core/NuGet.Packaging/Signing/Timestamp/Rfc3161TimestampProvider.cs)) and Hash Algorithm Name can be configured.
Automatically remove existing signature.

image

campersau:
There are 3 different resources: `NuGetPackageExplorer.Resources`, `PackageViewModel.Resources` and `Core.NuGetResources`.

In order to reuse as much as possible I have added the new strings into the `PackageViewModel.Resources` and made it public.

Is this okay?

PackageViewModel/Resources.Designer.cs
```diff
@@ -1,10 +1,10 @@
 ﻿//------------------------------------------------------------------------------
 // <auto-generated>
-//     This code was generated by a tool.
-//     Runtime Version:4.0.30319.42000
+//     Dieser Code wurde von einem Tool generiert.
```
campersau: All comments are german now. Thanks Visual Studio...

onovotny: lol

onovotny:
As you changed the title to remove WIP, does that mean this is ready for a final review & merge?

campersau:
@onovotny Yes, it is ready to merge.

onovotny:
Doing a quick test on it, it doesn't appear to be doing the timestamp. It's not remembering the setting for me either, so perhaps there's a bug transferring the value from the dialog to the settings?

onovotny:
Correction, it is remembering the chosen cert from the store, just that it's not transferring the value from the timestamp textbox to the view model.

Can you please fix that? Otherwise, it looks really good, thanks!

campersau:
Fixed.

onovotny:
Thanks!

Issue #359, opened 2018-02-23
file src attribute being overwritten


markjerz:
Hi! Thanks for reporting this feature/bug/question! 

Please keep / fill in the relevant info from this template so that we can help you as best as possible.

**Type** (choose one):

- Bug

**NPE version**: 4.1.32+gafe915819a

**Installed from**: Windows Store

In case of a BUG:

- What is the current result?

When using "Save Metadata As" it's overwriting the src attribute on files to match the target instead of leaving it as is. See the attached git diff:

![image](https://user-images.githubusercontent.com/3584066/36591857-3d4156e8-188b-11e8-9902-f3da6f68823b.png)

- What is the expected result?

It shouldn't be overwriting the src attribute. 

- Please post full exception details in case of an Exception (message, stacktrace, inner exceptions).
- Are there any workarounds? yes/no
- Is there a version in which it did work?

Yes, definitely the 3.x series, so feels like a recent regression.

304NotModified:
yes, unfortunately the nuspec will be fully generated from the specs 

This is also an issue if there is a wildcard in the src=

Bhaal22:
Hi,

I get this issue also while upgrading to 4.1.38.
3.25 is working fine

Bhaal22:
Issue is related to this Extension when ExportingManifest.
The cast is not possible and returns always null.

Will make a small PR for review. Not sure about the fix.

```c#
using NuGet.Packaging;

namespace NuGetPe
{
    public static class PackageFileExtensions
    {
        public static string OriginalPath(this IPackageFile packageFile)
        {
            return (packageFile as PhysicalPackageFile)?.SourcePath;
        }
    }
}
```

onovotny:
@Bhaal22 not sure what you mean by the cast not being possible?

Bhaal22:
@onovotny Just did a dummy PR

https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/pull/373

onovotny:
@Bhaal22 That's not correct, f is not ManifestFile, it's the input from the RootFolder.GetFiles(). They are IPackageFile and can be a PhysicalPackageFile.

Bhaal22:
I definitely do not say f type is ManifestFile

From what I have debugged, 

GetFiles() call returns List<IPackageFile> whose each element is of type PackageFile
and a PackageFile is not convertible to a PhysicalPackageFile. They both implements IPackageFile

[...]

PR #373, opened 2018-02-27
  #359

Bhaal22:
When iterating over GetFiles, the iterator type is PackageFile not PhysicalPackageFile.
Only f::_file is of type PhysicalPackageFile

How to reproduce the issue:

```
cd c:\tmp
mkdir nuget\binDebug

touch nuget\binDebug\file.txt

cp MyPackage.1.0.0.zip c:\tmp\nuget
unzip MyPackage.1.0.0.zip 
```

Now, open the nuspec with the NugetExplorer 4.x and Save the metadata.

The files xml section moved from:

```
<file src="binDebug\file.txt" target="content\file.txt" />
```

to

```
<file src="content\file.txt" target="content\file.txt" />
```

then the package does not open anymore or nuget pack complains because the src does not match the directory hierarchy.

[MyPackage 1.0.0.zip](https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/files/1763826/MyPackage.1.0.0.zip)


onovotny:
Can you please provide a repro...a before/after so I can see the difference?

Bhaal22:
@onovotny did you reproduce the issue?

onovotny:
Yes, I can see it. Question though, can the `as` ever return null; should it be using a null check? And If the extension method is no longer used (which it isn't in this case), can you please remove it?

Bhaal22:
I think the PR is not fully reliable. I will check to make to more "generic". I come back to you asap.

c0pp3rt0p:
I don't think this PR is going to fix the issue.  The underlying concrete type is already a PackageFile type in my testing so explicity casting to it does nothing.  The extension method being used is trying to convert to a PackageFileBase which PackageFile is not a descendant of so the conversion results in null.

Bhaal22:
The initial parent commit is:
d9d37f8ff097299b11c0a2e0ffb966d7d14d8d72

@c0pp3rt0p  I just doubler checked. The PR was definitely working when it has been done. 
With last source code version, indeed it does not work anymore.

c0pp3rt0p:
@Bhall22 Cheers, thanks for double checking.

c0pp3rt0p:
My hack to get through the day...  Moved the IEditablePackageFile interface to the Types project as well.

`
using NuGet.Packaging;

namespace NuGetPe
{
    public static class PackageFileExtensions
    {
        public static string OriginalPath(this IPackageFile packageFile)
        {
            if (packageFile is IEditablePackageFile)
                return (packageFile as IEditablePackageFile)?.OriginalPath;
            else
                return (packageFile as PackageFileBase)?.OriginalPath;
        }
    }
}
`

onovotny:
Closing due to age. Please re-open if still applicable and confirmed working.

PR #375, opened 2018-03-05
separate package name and version by "." in file names

### LINKED ISSUE #374: Saved packages no longer recognised by Local NuGet V2 Parser
**Type** : Bug

**NPE version**: 4.1.38

#361 broke all save methods because they also use the `ToString` method and the packages are no longer recognized by the [Local NuGet V2 Parser](https://github.com/NuGet/NuGet.Client/blob/02f95f1d5679ffe1063fcfd3d2508e91a2d881e6/src/NuGet.Core/NuGet.Protocol/Utility/LocalFolderUtility.cs#L534-L550).

---

https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/blob/d9d37f8ff097299b11c0a2e0ffb966d7d14d8d72/PackageViewModel/Commands/SavePackageCommand.cs#L121

https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/blob/d9d37f8ff097299b11c0a2e0ffb966d7d14d8d72/PackageViewModel/Commands/SavePackageCommand.cs#L145

https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/blob/d9d37f8ff097299b11c0a2e0ffb966d7d14d8d72/PackageViewModel/Commands/SavePackageCommand.cs#L182

https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/blob/d9d37f8ff097299b11c0a2e0ffb966d7d14d8d72/PackageViewModel/Commands/SavePackageCommand.cs#L213

https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/blob/d9d37f8ff097299b11c0a2e0ffb966d7d14d8d72/PackageViewModel/PackageViewModel.cs#L967

campersau:
`FileName` property name might be missleading. Any other ideas?

fixes #374

### DIFF
--- a/PackageViewModel/EditablePackageMetadata.cs
+++ b/PackageViewModel/EditablePackageMetadata.cs
@@ -587,6 +587,8 @@ namespace PackageExplorerViewModel
             return Id + " " + ManifestUtility.ReplaceMetadataWithToken(Version.ToFullString());
         }
 
+        public string FileName => Id + "." + ManifestUtility.ReplaceMetadataWithToken(Version.ToFullString());
+
         private string IsValid(string propertyName)
         {
             if (propertyName == "LicenseUrl")
--- a/PackageViewModel/EditablePackageMetadata.cs
+++ b/PackageViewModel/EditablePackageMetadata.cs
@@ -587,6 +587,9 @@ namespace PackageExplorerViewModel
             return Id + " " + ManifestUtility.ReplaceMetadataWithToken(Version.ToFullString());
         }
 
+        /// <summary>
+        /// file name for V2 local feeds {id}.{version}
+        /// </summary>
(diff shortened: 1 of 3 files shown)


PackageViewModel/EditablePackageMetadata.cs
```diff
@@ -587,6 +587,8 @@ public override string ToString()
             return Id + " " + ManifestUtility.ReplaceMetadataWithToken(Version.ToFullString());
         }
 
+        public string FileName => Id + "." + ManifestUtility.ReplaceMetadataWithToken(Version.ToFullString());
```
304NotModified: maybe we should add a comment here that the `.` is required?

onovotny:
I'm not crazy about relying on ToString being used for "business logic". I'd rather see a new method that returns a string that's used for a specific purpose.

onovotny:
I like the extra properties for this

Issue #383, opened 2018-03-24
Support drag & drop of files to outside of NPE


svick:
**Type**: Feature request

**NPE version**: 4.2.13

**Installed from**: Windows Store

I wanted to extract a file from a NuGet package, so I did the obvious thing: open the package in NPE and then drag and drop the file from NPE to a folder. But it didn't work, the mouse cursor turns into the "you can't drop here" state.

Pretty much the same thing can be accomplished by selecting "Save As…" from the context menu, so this is just a convenience feature.

campersau:
I am currently working on it. Also going to implement copy & past (both ways) (maybe also with context menu).
This will only work for single files (not folders) for now.

PR #384, opened 2018-03-24
support drag & drop outside of NPE and implement copy & paste

campersau:
fixes https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/issues/383

It is now also possible to copy `PackageFolder`s inside of NPE.

### DIFF
--- a/PackageExplorer/Interop/NativeDragDrop.cs
+++ b/PackageExplorer/Interop/NativeDragDrop.cs
@@ -58,14 +58,18 @@ namespace PackageExplorer
 
             for (var i = 0; i < fileNames.Length; i++)
             {
-                var stream = GetStream(windowsDataObject, i);
+                Stream stream = null;
+                if (!fileNames[i].Value) // not a directory
+                {
+                    stream = GetStream(windowsDataObject, i);
+                }
 
-                yield return new KeyValuePair<string, Stream>(fileNames[i], stream);
+                yield return new KeyValuePair<string, Stream>(fileNames[i].Key, stream);
             }
         }
 
         // https://stackoverflow.com/questions/8709076/drag-and-drop-multiple-attached-file-from-outlook-to-c-sharp-window-form
-        private static string[] GetFileGroupDescriptorWFileNames(WindowsIDataObject data)
+        private static KeyValuePair<string, bool>[] GetFileGroupDescriptorWFileNames(WindowsIDataObject data)
         {
             var fileGroupDescriptorWPointer = IntPtr.Zero;
             try
@@ -81,7 +85,7 @@ namespace PackageExplorer
                 var fileGroupDescriptor = Marshal.PtrToStructure<FILEGROUPDESCRIPTORW>(fileGroupDescriptorWPointer);
 
                 //create a new array to store file names in of the number of items in the file group descriptor
-                var fileNames = new string[fileGroupDescriptor.cItems];
+                var fileNames = new KeyValuePair<string, bool>[fileGroupDescriptor.cItems];
 
                 //get the pointer to the first file descriptor
                 var fileDescriptorPointer = (IntPtr)((long)fileGroupDescriptorWPointer + Marshal.SizeOf(fileGroupDescriptor.cItems));
@@ -91,7 +95,17 @@ namespace PackageExplorer
                 {
                     //marshal the pointer top the file descriptor as a FILEDESCRIPTORW struct and get the file name
                     var fileDescriptor = Marshal.PtrToStructure<FILEDESCRIPTORW>(fileDescriptorPointer);
-                    fileNames[fileDescriptorIndex] = fileDescriptor.cFileName;
+
+                    var isDirectory = false;
+                    if ((fileDescriptor.dwFlags & FD_ATTRIBUTES) == FD_ATTRIBUTES)
+                    {
+                        if ((fileDescriptor.dwFileAttributes & FILE_ATTRIBUTE_DIRECTORY) == FILE_ATTRIBUTE_DIRECTORY)
+                        {
+                            isDirectory = true;
+                        }
+                    }
+
+                    fileNames[fileDescriptorIndex] = new KeyValuePair<string, bool>( fileDescriptor.cFileName, isDirectory);
 
                     //move the file descriptor pointer to the next file descriptor
                     fileDescriptorPointer = (IntPtr)((long)fileDescriptorPointer + Marshal.SizeOf(fileDescriptor));
@@ -280,7 +294,11 @@ namespace PackageExplorer
         }
 
         // https://msdn.microsoft.com/en-us/library/windows/desktop/bb773288(v=vs.85).aspx
+        private const uint FD_ATTRIBUTES = 0x00000004;
         private const uint FD_SHOWPROGRESSUI = 0x00004000;
+
+        // https://msdn.microsoft.com/en-us/library/windows/desktop/gg258117(v=vs.85).aspx
+        private const int FILE_ATTRIBUTE_DIRECTORY = 0x10;
 
         // https://msdn.microsoft.com/en-us/library/windows/desktop/bb773290(v=vs.85).aspx
         [StructLayout(LayoutKind.Sequential)]
--- a/PackageExplorer/Interop/NativeDragDrop.cs
+++ b/PackageExplorer/Interop/NativeDragDrop.cs
@@ -76,7 +76,7 @@ namespace PackageExplorer
             ref var fileGroupDescriptorPtr = ref MemoryMarshal.GetReference(fileGroupDescriptorBytes);
             var fileGroupDescriptor = Unsafe.As<byte, FILEGROUPDESCRIPTORW>(ref fileGroupDescriptorPtr);
 
-            var fileNames = new (string, bool)[fileGroupDescriptor.cItems];
+            var fileNames = new(string, bool)[fileGroupDescriptor.cItems];
             unsafe
             {
                 fixed (byte* pStart = &Unsafe.Add(ref fileGroupDescriptorPtr, Marshal.SizeOf<FILEGROUPDESCRIPTORW>()))
@@ -204,39 +204,29 @@ namespace PackageExplorer
                 _inner.Commit(0); // https://msdn.microsoft.com/en-us/library/windows/desktop/aa380320%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396
             }
 
-            public override int Read(byte[] buffer, int offset, int count)
+            public override unsafe int Read(byte[] buffer, int offset, int count)
             {
-                var ptr = Marshal.AllocHGlobal(Marshal.SizeOf<ulong>());
-                try
+                if (offset > 0)
                 {
-                    if (offset > 0)
-                    {
-                        buffer = buffer.Skip(offset).ToArray();
-                    }
-
-                    _inner.Read(buffer, count, ptr);
-
-                    return (int)Marshal.ReadInt64(ptr);
+                    buffer = buffer.Skip(offset).ToArray();
                 }
-                finally
-                {
-                    Marshal.FreeHGlobal(ptr);
-                }
+
+                ulong lng = 0;
+                var p = &lng;
+
+                _inner.Read(buffer, count, (IntPtr)p);
+
+                return (int)lng;
             }
 
-            public override long Seek(long offset, SeekOrigin origin)
+            public override unsafe long Seek(long offset, SeekOrigin origin)
             {
-                var ptr = Marshal.AllocHGlobal(Marshal.SizeOf<ulong>());
-                try
-                {
-                    _inner.Seek(offset, (int)origin, ptr);
+                long lng = 0;
+                var p = &lng;
 
-                    return Marshal.ReadInt64(ptr);
-                }
-                finally
-                {
-                    Marshal.FreeHGlobal(ptr);
-                }
+                _inner.Seek(offset, (int)origin, (IntPtr)p);
+
+                return lng;
             }
 
             public override void SetLength(long value)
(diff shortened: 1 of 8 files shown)


PackageExplorer/PackageViewer.xaml.cs
```diff
@@ -489,13 +470,109 @@ private void PackageFolderContextMenu_Opened(object sender, RoutedEventArgs e)
...
+                    return true;
+                }
+            }
+            return false;
+        }
+
+        private IDataObject CreateDataObject(PackagePart packagePart)
+        {
+            var data = new DataObject();
+            data.SetData(PackageFileDataFormat, packagePart.Path);
```
campersau: During drag & drop the object does not have to be `Serializable` but for copy & past it is required.
That's why now the `PackageFileDataFormat` is just the path and not the actual `PackagePart`.
Since drag & drop from one NPE instance into another one didn't work before this is not a breaking change.

PackageExplorer/PackageViewer.xaml.cs
```diff
@@ -489,13 +470,109 @@ private void PackageFolderContextMenu_Opened(object sender, RoutedEventArgs e)
...
+
+            if (data.GetDataPresent(PackageFileDataFormat))
+            {
+                if (data.GetData(PackageFileDataFormat) is string packagePartPath)
+                {
+                    folder = folder ?? RootFolder;
+
+                    var packagePart = RootFolder.GetPackageParts().FirstOrDefault(part => part.Path == packagePartPath);
+
+                    if (packagePart != null)
```
campersau: This and other things currently prevent drag & drop / copy & past from one NPE instance into another one.

onovotny:
LGTM, lmk when you're done or if you have more changes you need in this PR

campersau:
I also added copy & paste menu items. Files can be copied. In folders can be pasted.

image

image

It also supports copying from other windows applications like RDP / Outlook via IStream.

campersau:
@onovotny ready for merge!

onovotny:
Is there a place for the new `Span<T>` or `Memory<T>` in any of the interop code?

campersau:
@onovotny Yes, it looks like `Marshal.Copy` can be replaced with `Span`s and `Unsafe.ToPointer` but then we have to compile with `/unsafe`. See https://github.com/campersau/NuGetPackageExplorer/commit/37039787c72056f2e9dc6d743292003001ebb524?w=1

benaadams:
> and `Unsafe.ToPointer` but then we have to compile with /unsafe

Would also need to fix the span or array over the `Unsafe.AsPointer` + `Marshal.StructureToPtr` call; else you have a gc hole and the `IntPtr` may not point to the correct location for the `StructureToPtr` call

campersau:
@benaadams Do you mean something like this? Also if we need to fix them, we could just do it directly without `Span`s:

Without Span
```csharp
unsafe
{
    fixed (byte* p = bytes)
    {
        Marshal.StructureToPtr(source, (IntPtr)p, false);
    }
}
```

With Span
```csharp
unsafe
{
    fixed (byte* p = bytes)
    {
        MarshalStructure(new ReadOnlySpan<byte>(p, bytes.Length));
    }
}

return bytes;

unsafe void MarshalStructure(ReadOnlySpan<byte> fixedBytes)
{
    var p = Unsafe.AsPointer(ref MemoryMarshal.GetReference(fixedBytes));

    Marshal.StructureToPtr(source, (IntPtr)p, false);
}
````

---

Without Span
```csharp
unsafe
{
    fixed (byte* p = fileGroupDescriptorBytes)
    {
        return MarshalFileNames(p);
    }
}

unsafe string[] MarshalFileNames(byte* fixedBytes)
{
    var fileGroupDescriptor = Marshal.PtrToStructure<FILEGROUPDESCRIPTORW>((IntPtr)fixedBytes);
    var fileNames = new string[fileGroupDescriptor.cItems];

    fixedBytes = fixedBytes + Marshal.SizeOf<FILEGROUPDESCRIPTORW>();

    for (var fileDescriptorIndex = 0; fileDescriptorIndex < fileGroupDescriptor.cItems; fileDescriptorIndex++)
    {
        var fileDescriptor = Marshal.PtrToStructure<FILEDESCRIPTORW>((IntPtr)fixedBytes);
        fileNames[fileDescriptorIndex] = fileDescriptor.cFileName;

        fixedBytes = fixedBytes + Marshal.SizeOf<FILEDESCRIPTORW>();
    }
    return fileNames;
}
```

With Span
```csharp
unsafe
{
    fixed (byte* p = fileGroupDescriptorBytes)
    {
        return MarshalFileNames(new ReadOnlySpan<byte>(p, fileGroupDescriptorBytes.Length));
    }
}

unsafe string[] MarshalFileNames(ReadOnlySpan<byte> fixedBytes)
{
    var p = Unsafe.AsPointer(ref MemoryMarshal.GetReference(fixedBytes));

    var fileGroupDescriptor = Marshal.PtrToStructure<FILEGROUPDESCRIPTORW>((IntPtr)p);
    var fileNames = new string[fileGroupDescriptor.cItems];

    p = Unsafe.Add<byte>(p, Marshal.SizeOf<FILEGROUPDESCRIPTORW>());

    for (var fileDescriptorIndex = 0; fileDescriptorIndex < fileGroupDescriptor.cItems; fileDescriptorIndex++)
    {
        var fileDescriptor = Marshal.PtrToStructure<FILEDESCRIPTORW>((IntPtr)p);
        fileNames[fileDescriptorIndex] = fileDescriptor.cFileName;

        p = Unsafe.Add<byte>(p, Marshal.SizeOf<FILEDESCRIPTORW>());
    }
    return fileNames;
}
```

benaadams:
Can just fix the return from `MemoryMarshal.GetReference` and keep it all in the one method:
```csharp
unsafe void MarshalStructure(ReadOnlySpan<byte> bytes)
{
    fixed (byte *p = &MemoryMarshal.GetReference(bytes))
    {
        Marshal.StructureToPtr(source, (IntPtr)p, false);
    }
}
```
The return from `Unsafe.AsPointer` can be moved after the call if its not already fixed so become incorrect. `fixed` is close to zero cost *if* no gc happens during the fixed time.

benaadams:
Though if you don't have anything fancy going on in the struct (e.g. `FILEGROUPDESCRIPTORW` doesn't look exciting) could use `Unsafe.As`

`FILEDESCRIPTORW` on the other hand is an exciting struct as it has a `string` in it so you would still need to marshal that

benaadams:
Something like?
```csharp
unsafe string[] MarshalFileNames(ReadOnlySpan<byte> bytes)
{
    ref var data = ref MemoryMarshal.GetReference(bytes);
    var fileGroupDescriptor = Unsafe.As<byte, FILEGROUPDESCRIPTORW>(ref data);
    var fileNames = new string[fileGroupDescriptor.cItems];

    fixed (byte* pStart = &Unsafe.Add(ref data, Marshal.SizeOf<FILEGROUPDESCRIPTORW>()))
    {
        var ptr = pStart;

        for (var fileDescriptorIndex = 0; fileDescriptorIndex < fileGroupDescriptor.cItems; fileDescriptorIndex++)
        {
            var fileDescriptor = Marshal.PtrToStructure<FILEDESCRIPTORW>((IntPtr)ptr);
            fileNames[fileDescriptorIndex] = fileDescriptor.cFileName;

            ptr += Marshal.SizeOf<FILEDESCRIPTORW>();
        }
    }

    return fileNames;
}
```

campersau:
Thanks @benaadams !
Here is your suggestion https://github.com/campersau/NuGetPackageExplorer/commit/1a3bccddd8790a71f8d8cbdb12f46348775e61d9?diff=split&w=1

Let's see what @onovotny thinks about the 3 alternatives and if this is worth the additional dependencies & unsafe code.

onovotny:
TBH, I don't have any strong feeling either way. I'm okay with using `unsafe`, but beyond that, I'm good with whatever makes the most sense for readability and reducing potential errors in the code. I don't have a strong sense of what the best approach is in various circumstances yet, hence the question :)

What do you both think?

benaadams:
> Here is your suggestion campersau/NuGetPackageExplorer@1a3bccd?diff=split&w=1

You can drop the outer `unsafe`/`fixed`; also make use of implicit conversion to ReadOnlySpan so
```csharp
var fileGroupDescriptorStream = (MemoryStream)data.GetData(FileGroupDescriptorW);
var fileGroupDescriptorBytes = fileGroupDescriptorStream.ToArray();
 	 
unsafe
{
    fixed (byte* p = fileGroupDescriptorBytes)
    {
        return MarshalFileNames(new ReadOnlySpan<byte>(p, fileGroupDescriptorBytes.Length));
    }
}
```
Could become:
```csharp
var fileGroupDescriptorStream = (MemoryStream)data.GetData(FileGroupDescriptorW);
var fileGroupDescriptorBytes = fileGroupDescriptorStream.ToArray();
return MarshalFileNames(fileGroupDescriptorBytes);
```

benaadams:
> What do you both think?

Personally I prefer staying on the managed heap rather than using `Marshal.AllocHGlobal` + `Marshal.FreeHGlobal`; but I have an aversion to manually deallocating memory 😄

onovotny:
@benaadams I agree, I like the managed heap much better, in general.

campersau:
@benaadams So fixing is only required if we are dealing with pointers and not with refs?

Then we also don't need the local function right?

```csharp
ReadOnlySpan<byte> fileGroupDescriptorBytes = fileGroupDescriptorStream.ToArray();

ref var fileGroupDescriptorPtr = ref MemoryMarshal.GetReference(fileGroupDescriptorBytes);
var fileGroupDescriptor = Unsafe.As<byte, FILEGROUPDESCRIPTORW>(ref fileGroupDescriptorPtr);

var fileNames = new string[fileGroupDescriptor.cItems];
unsafe
{
    fixed (byte* pStart = &Unsafe.Add(ref fileGroupDescriptorPtr, Marshal.SizeOf<FILEGROUPDESCRIPTORW>()))
    {
        var fileDescriptorRowPtr = pStart;
        for (var fileDescriptorIndex = 0; fileDescriptorIndex < fileGroupDescriptor.cItems; fileDescriptorIndex++)
        {
            var fileDescriptor = Marshal.PtrToStructure<FILEDESCRIPTORW>((IntPtr)fileDescriptorRowPtr);
            fileNames[fileDescriptorIndex] = fileDescriptor.cFileName;

            fileDescriptorRowPtr += Marshal.SizeOf<FILEDESCRIPTORW>();
        }
    }
}
return fileNames;
```

benaadams:
> So fixing is only required if we are dealing with pointers and not with refs?

Yeah, the GC will track the `ref`s and update them if it moves memory, also is aware they are in use. 

As far as the runtime is concerned pointers and `IntPtr` are just numbers so have no coupling to the actual memory; so you need to either fix the memory, use a pinned GC handle or use native allocation (which is fixed) to stop it potentially moving.

PackageExplorer/Interop/NativeDragDrop.cs
```diff
@@ -0,0 +1,315 @@
...
+                }
+                finally
+                {
+                    Marshal.FreeHGlobal(ptr);
+                }
+            }
+
+            public override long Seek(long offset, SeekOrigin origin)
+            {
+                var ptr = Marshal.AllocHGlobal(Marshal.SizeOf<ulong>());
```
benaadams: Since you are going unsafe... 😉

This could just use stack memory (stack is also fixed memory): 
```csharp
public unsafe override long Seek(long offset, SeekOrigin origin)
{
    long result = 0L;
    var ptr = &result;

    _inner.Seek(offset, (int)origin, (IntPtr)ptr);

    return result;
}
```

PackageExplorer/Interop/NativeDragDrop.cs
```diff
@@ -0,0 +1,315 @@
...
+            }
+
+            public override void Flush()
+            {
+                _inner.Commit(0); // https://msdn.microsoft.com/en-us/library/windows/desktop/aa380320%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396
+            }
+
+            public override int Read(byte[] buffer, int offset, int count)
+            {
+                var ptr = Marshal.AllocHGlobal(Marshal.SizeOf<ulong>());
```
benaadams: Could also use stack here instead
```csharp
public unsafe override int Read(byte[] buffer, int offset, int count)
{
    long result = 0L;
    var ptr = &result;

    if (offset > 0)
    {
        buffer = buffer.Skip(offset).ToArray();
    }

    _inner.Read(buffer, count, (IntPtr)ptr);

    return (int)result;
}
```

PackageExplorer/Interop/NativeDragDrop.cs
```diff
@@ -0,0 +1,315 @@
...
+            }
+
+            public override void SetLength(long value)
+            {
+                _inner.SetSize(value);
+            }
+
+            public override void Write(byte[] buffer, int offset, int count)
+            {
+                var ptr = Marshal.AllocHGlobal(Marshal.SizeOf<ulong>());
```
benaadams: Also could use stack here
```csharp
public unsafe override void Write(byte[] buffer, int offset, int count)
{
    long result = 0L;
    var ptr = &result;

    if (offset > 0)
    {
        buffer = buffer.Skip(offset).ToArray();
    }

    _inner.Write(buffer, count, (IntPtr)ptr);

    var written = (int)result;

    if (written != count)
    {
        Write(buffer, written, count - written);
    }
}
```

campersau:
@benaadams Thanks for the clarification!

@onovotny I updated the code to use the managed heap instead. You can merge if you are ok with it.

PackageExplorer/Interop/NativeDragDrop.cs
```diff
@@ -204,68 +204,51 @@ public override void Flush()
...
                 {
-                    if (offset > 0)
-                    {
-                        buffer = buffer.Skip(offset).ToArray();
-                    }
+                    buffer = buffer.Skip(offset).ToArray();
+                }
 
-                    _inner.Write(buffer, count, ptr);
+                var written = 0;
```
benaadams: Should this be a `long` rather than `int`? Don't know the api that well

campersau: In the docs I coun't find anything about the type https://msdn.microsoft.com/en-us/library/system.runtime.interopservices.comtypes.istream.write(v=vs.110).aspx

But I thought that if the `count` is an `int` then the result can't be bigger.

benaadams: Yeah docs for Read has **`ULONG`** in bold for the return type https://msdn.microsoft.com/en-us/library/system.runtime.interopservices.comtypes.istream.read(v=vs.110).aspx

Suppose if it works 😄

benaadams: Ah it says refer to `ISequentialStream::Write` which does say its a pointer to a `ulong` https://msdn.microsoft.com/en-us/library/windows/desktop/aa380014(v=vs.85).aspx

However as the other param is also a ulong its likely a C ulong; meaning they are all C# uint/int sizes. Same for the other methods

onovotny:
Thanks! merging this in now, will wait a little bit before releasing to the store/choco in case there's stuff that creeps in.

PR #387, opened 2018-04-09
check if ContentType header exits

### LINKED ISSUE #386: AppVeyor remote feed package open fails with "Object reference not set to an instance of an object" message

**Type**:

- Bug

**NPE version**: 4.2.13+g3c913990f30

**Installed from**: chocolatey.org  and Windows Store (tried both, currently using Windows Store installed version)

In case of a BUG:

- What is the current result?
Every time I try to open a package from our private AppVeyor feed the package fails to either open or download with an "Object reference not set to an instance of an object" exception. 

Packages and versions list correctly; they simply fail to open.

Package should open or download as it does for NuGet.org packages.

- Please post full exception details in case of an Exception (message, stacktrace, inner exceptions).

"Object reference not set to an instance of an object"

- Are there any workarounds? yes/no

Can retrieve package via other methods then open as local package.

- Is there a version in which it did work?

Don't know.

Not that I am aware of.

image

campersau:
e.g. AppVeyor does not send it

fixes #386

### DIFF
--- a/PackageExplorer/MefServices/PackageDownloader.cs
+++ b/PackageExplorer/MefServices/PackageDownloader.cs
@@ -213,7 +213,7 @@ namespace PackageExplorer
         {
             var response = await base.SendAsync(request, cancellationToken);
 
-            if (IsBinaryMediaType(response.Content.Headers.ContentType.MediaType))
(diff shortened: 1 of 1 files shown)


304NotModified:
Thanks!

onovotny:
thanks!

PR #389, opened 2018-04-09
reinitialize credential service after new credentials are added

### LINKED ISSUE #388: Credentials are not cached
**Type**: Bug

**NPE version**: master

**Installed from**: compiled

- What is the current result?
Credentials are not cached

Credentials should be cached

- Are there any workarounds?
no

- Is there a version in which it did work?
never

You need a private feed with authentification

---

Search for a package in a private feed and download it.
NPE should only ask for credentials once. But it currently does it twice: searching, downloading.

---

The problem is, that the `CredentialService` caches all [responses](https://github.com/NuGet/NuGet.Client/blob/d873418a3a3188e405c5815e997115b1fda73cd9/src/NuGet.Clients/NuGet.Credentials/CredentialService.cs#L21-L22) from the `ICredentialProvider`s and never clears them.

And new credentials are added here at runtime:

https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/blob/f47259ce5273b5d8ddf7a903c4950c1e24c56027/PackageExplorer/App.xaml.cs#L47

campersau:
because the CredentialService is caching the responses to detect retries
(it would be nicer if we could just clear the retry cache)

fixes #388

### DIFF
--- a/PackageExplorer/App.xaml.cs
+++ b/PackageExplorer/App.xaml.cs
@@ -39,12 +39,12 @@ namespace PackageExplorer
 
         private async void Application_Startup(object sender, StartupEventArgs e)
         {
-            HttpHandlerResourceV3.CredentialService = new CredentialService(new ICredentialProvider[] {
-                Container.GetExportedValue<CredentialManagerProvider>(),
-                Container.GetExportedValue<CredentialPublishProvider>(),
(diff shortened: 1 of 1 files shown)


onovotny:
Thanks!

PR #392, opened 2018-04-10
check if lastWriteTime is valid

campersau:
otherwise an exception is thrown

```
System.ArgumentOutOfRangeException
  HResult=0x80131502
  Message=Ungültige Win32-FileTime.
  Source=mscorlib
  StackTrace:
   at System.DateTime.ToFileTimeUtc()
```

### DIFF
--- a/Core/Packages/DiskPackageFile.cs
+++ b/Core/Packages/DiskPackageFile.cs
@@ -1,0 +1,40 @@
+﻿using System;
+using System.IO;
+
+namespace NuGetPe
+{
+    public class DiskPackageFile : PackageFileBase
+    {
+        private DateTimeOffset? _lastWriteTime;
+
+        public DiskPackageFile(string path, string physicalPath)
+            : base(path)
+        {
+            OriginalPath = physicalPath;
+        }
+
+        public override Stream GetStream()
+        {
+            return File.OpenRead(OriginalPath);
+        }
+
+        public override string OriginalPath { get; }
+
+        public override DateTimeOffset LastWriteTime
+        {
+            get
+            {
+                if (!_lastWriteTime.HasValue)
+                {
+                    _lastWriteTime = File.GetLastWriteTimeUtc(OriginalPath);
+                }
+                return _lastWriteTime.Value;
+            }
+        }
+
+        public override string ToString()
+        {
+            return Path;
+        }
+    }
+}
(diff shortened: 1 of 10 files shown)


PackageExplorer/Interop/NativeDragDrop.cs
```diff
@@ -22,15 +22,18 @@ public static Stream CreateFileGroupDescriptorW(string fileName, DateTimeOffset
...
-            var changeTimeFileTime = new System.Runtime.InteropServices.ComTypes.FILETIME
+            if (lastWriteTime != default(DateTimeOffset))
             {
-                dwLowDateTime = (int)(changeTime & 0xffffffff),
-                dwHighDateTime = (int)(changeTime >> 32),
-            };
-            fileDescriptor.ftLastWriteTime = changeTimeFileTime;
-            fileDescriptor.ftCreationTime = changeTimeFileTime;
+                fileDescriptor.dwFlags |= FD_CREATETIME | FD_WRITESTIME;
+                var changeTime = lastWriteTime.ToLocalTime().ToFileTime();
```
onovotny: Should this be local time or UTC for this API?

campersau: That's some of the code I copied from http://dlaa.me/blog/post/9917797.

But looking at the source code of `DateTimeOffset.ToFileTime()` the call to `ToLocalTime()` is acutally unnecessary.

https://github.com/dotnet/corefx/blob/aeda6ab039f8d6e6041379ab156d51410b679c13/src/Common/src/CoreLib/System/DateTimeOffset.cs#L711-L714
or
https://referencesource.microsoft.com/#mscorlib/system/datetimeoffset.cs,621

I will remove it.

campersau:
Hmm.. the main reason why the last write time is not set in the first place is that the `PhysicalPackageFile` reads the last write date only if the [stream is accessed](https://github.com/NuGet/NuGet.Client/blob/90524ca5f2313702df2f29b72649c9611aa4b086/src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/PhysicalPackageFile.cs#L73-L103).

So the actual fix would be to create a wrapper or just implement our own `PhysicalPackageFile` which reads the last write date if needed an not just when the stream is accessed.

onovotny:
Either sounds good, whichever approach is simplest.

Core/Packages/DiskPackageFile.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System;
+using System.IO;
+
+namespace NuGetPe
+{
+    public class DiskPackageFile : PackageFileBase
```
campersau: I wanted to avoid naming collisions... Other suggestions?

PR #394, opened 2018-04-10
WAP references nuget package dlls from ref instead of lib folder

### LINKED ISSUE #391: BadImageFormatException with drag/drop
I'm getting an exception if I try to drag/drop a file. I'm on RS4 (17133):

image

@campersau 

Looks like if I use NuGetPackageExplorer as the startup project, it works, but if I use PackageExplorer.Package.Nightly, it fails. 

Seems like the wrong assembly is being deployed for System.Memory.

campersau:
This is a pretty hacky attempt to fix #391

During the WAP build process the `_GenerateAppxPackageRecipe` target creates a xml file called `$(TargetName).build.appxreceipe` which contains all dlls referenced by the `ProjectReference`s.
Because WAP somehow prefers dlls from the `ref` folder instead of the `lib` folder we can just replace the folder names.

Todo:
- [x] apply this fix to `PackageExplorer.Package`

### DIFF
--- a/PackageExplorer.Package.Nightly/PackageExplorer.Package.Nightly.wapproj
+++ b/PackageExplorer.Package.Nightly/PackageExplorer.Package.Nightly.wapproj
@@ -63,6 +63,7 @@
   <ItemGroup>
     <ProjectReference Include="..\PackageExplorer\NuGetPackageExplorer.csproj" SkipGetTargetFrameworkProperties="true" />
   </ItemGroup>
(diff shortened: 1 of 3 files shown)


onovotny:
@campersau I want to look closer at the binlog to see if I can prevent the reference from coming through at all

onovotny:
Lets go with this hack for now and we'll remove it once fixed in VS. Can you also apply to the main package?

PR #396, opened 2018-04-11
Change PrereleasePackageDependencyRule from Error to Warning

### LINKED ISSUE #395: PrereleasePackageDependencyRule rule required?
**Type** (choose one):

- Question

**NPE version**: (e.g. 4.3.12) (see Help -> about)

**Installed from**: chocolatey.org

<hr />

From what I gather from https://github.com/NuGet/NuGet.Client/pull/1073 , the Nuget client treats prerelease dependencies in stable packages as a warning now and not an error. Should NPE not do the same? So my question is: Is the `PrereleasePackageDependencyRule` validation rule still required?

If it's a simple case of removing the rule, I'll submit a PR.

igitur:
Fixes #395


onovotny:
Thanks!

PR #408, opened 2018-04-28
avoid number parsing if value is already a number

campersau:
The converter is currently only used for `DownloadCount` which is already an `int`.


onovotny:
cherry-picked in

PR #405, opened 2018-04-28
New Package Chooser Dialog

onovotny:
PR to review/merge in the new package chooser dialog that @khaister contributed.

Would appreciate a review from @campersau as well.

PR #411, opened 2018-04-28
use SelectedPackage which is updated during version selection

campersau:
This implements the "cool feature" from https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/pull/405#issuecomment-385183453 and shows the publish date if a package is selected from the version list.


onovotny:
cherry-picked in

PR #430, opened 2018-06-08
Recenter pkg chooser dialog to owner when shown (fixes #418)

### LINKED ISSUE #418: File -> Open from Feed opens the dialog in the previous location; should be center parent
**Type** (choose one):

- **Bug**
- ~Feature request~
- ~Question~

**NPE version**: "Edge" Version

**Installed from**: Built from Source

- *What is the current result?* The package chooser dialog shows up in the previously opened location.
- *What is the expected result?* The dialog should be center parent (`CenterOwner`).
- *Are there any workarounds?* Not that I know of.
- *Is there a version in which it did work?* Not that I know of.

khaister:


onovotny:
Thanks!

PR #437, opened 2018-06-14
make streams seekable before consuming them in ImageFileViewer / RtfF…

### LINKED ISSUE #432: Can't preview RTF file
**Type**: Bug

**NPE version**: 4.4.3

When I try to preview the ThirdPartyNotices.rtf file from [the Microsoft.Net.Compilers 2.8.2 package](https://www.nuget.org/packages/Microsoft.Net.Compilers/2.8.2), I get the following exception:

```
*** The plugin that registers to read this file type fails to read the content of this file. ***
System.NotSupportedException: This operation is not supported.
   at System.IO.Compression.DeflateStream.get_Length()
   at PackageExplorer.RtfFileViewer.GetView(String extension, Stream stream)
   at PackageExplorerViewModel.ViewContentCommand.ShowFile(PackageFile file)
```

I don't know anything about this code, but the obvious workaround would be copy the `DeflateStream` into a `MemoryStream` and use that.

campersau:
…ileViewer / FileEditor

fixes #432

### DIFF
--- a/PackageExplorer/Utilities/StreamUtility.cs
+++ b/PackageExplorer/Utilities/StreamUtility.cs
@@ -9,5 +9,17 @@ namespace PackageExplorer
         {
             return new MemoryStream(Encoding.UTF8.GetBytes(content));
         }
+
+        public static Stream MakeSeekable(Stream stream)
+        {
+            if (stream.CanSeek)
+            {
+                return stream;
+            }
+            var memoryStream = new MemoryStream();
+            stream.CopyTo(memoryStream);
+            memoryStream.Position = 0;
+            return memoryStream;
+        }
     }
 }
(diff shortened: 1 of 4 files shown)


PackageExplorer/Utilities/StreamUtility.cs
```diff
@@ -9,5 +9,17 @@ public static Stream ToStream(string content)
         {
             return new MemoryStream(Encoding.UTF8.GetBytes(content));
         }
+
+        public static Stream MakeSeekable(Stream stream)
```
campersau: The name comes from amazon 😄 
https://docs.aws.amazon.com/sdkfornet1/latest/apidocs/html/M_Amazon_S3_Util_AmazonS3Util_MakeStreamSeekable.htm

Issue #442, opened 2018-07-02
Package source URL with a trailing space fails


onovotny:
If you paste in a url for a feed source and leave a trailing space, loading fails.

Leading/trailing spaces should be trimmed.

kyungjaeson:
Newbie here, 
I am interested in taking on this issue

onovotny:
Yes, please fork and submit a PR.

kyungjaeson:
PR up

PR #444, opened 2018-07-04
Issue #442

kyungjaeson:
Fixed as per request


kyungjaeson:
So checking the docs it always returns a string. 
Double checked the application via debugger, and it's an empty string. 
issue442ngpegithub

onovotny:
Thanks!

Issue #434, opened 2018-06-13
Package file is locked while opened


ramonsmits:
**Type** (choose one):

- Bug / Feature request

**NPE version**: 4.4.3
**Installed from**: Windows Store

When NPE has a package open, it locks the nuget zip file. It cannot be overwritten by visual studio which means NPE must exit or close the opened package.

It would be very nice if it would not lock the file while open or even auto-refresh detecting the file is overwritten.

onovotny:
This has been a long-standing issue. NPE uses an open read/write in case the user makes a change and hits save.


I believe it should use an open read, and only use a read/write for any saves. 


PR's accepted :)

PR #458, opened 2018-09-01
dispose package streams

campersau:
Just found these in the code.


onovotny:
Thanks! Do you know if any of the outstanding issues are resolved by this?

campersau:
No, I don't think so, but it may be required for #434.

onovotny:
Sounds good. I'd really like to fix #434 but I haven't had time to dig in.

Issue #376, opened 2018-03-14
A local file header is corrupt


asiridissa:
Bug
4.138
Windows Store

It displays following error message on save
--------------------------- NuGet Package Explorer ---------------------------
 A local file header is corrupt.
--------------------------- OK ---------------------------

Had no problems with the previous version.

I just copied the previous version of .nupkg file and rename it to the new version and after editing the file, on save the problem occurs.

onovotny:
What package is this occurring with, can you attach it here (rename it to .zip)

asiridissa:
Here is a package with the issue.

[Robot.Dimension.Tenant 1.0.4.1.zip](https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/files/1833113/Robot.Dimension.Tenant.1.0.4.1.zip)

onovotny:
I just opened it on my machine without issue:

![image](https://user-images.githubusercontent.com/1427284/37708894-b602fec6-2cde-11e8-9402-81799bd698bc.png)

What version of .NET and what OS are you on?

asiridissa:
Im on Windows 10 Pro 1709 and .Net 4.7.02556.
Here is my last scenario.
1. Delete a file.
2. Change the version.
3. Ctrl+S (Saves fine).
Then Ctrl+S again will prompt the error.
If I select the publish menu, also error prompts.

![2](https://user-images.githubusercontent.com/3539388/37750475-4258b1d4-2db3-11e8-9f16-0437195b8736.PNG)

onovotny:
Thanks. I can repro now on save. Investigating.

mmoore99:
Any update on this issue? I am experiencing the same problem.

simonlaszcz:
Is there a workaround for this?

igitur:
I ran NPE from source and got this problem too. The exception occurs in `System.IO.Packaging`, so I think it's a bug there. Sorry, I don't have the stack trace anymore. Will try to reproduce later.

So what I did was repackage the `.nupkg` file manually (renamed to zip, extract, rezip, rename to .nupkg). That fixed my issue.

igitur:
Stack trace:
```
   at System.IO.Compression.ZipArchiveEntry.OpenInReadMode(Boolean checkOpenable)
   at NuGet.Packaging.PackageArchiveReader.GetStream(String path)
[...]

PR #460, opened 2018-09-01
do not lock package file if no changes are made

### LINKED ISSUE #434: Package file is locked while opened

campersau:
- If the users makes changes to the package without saving them the package will be locked to prevent deletion / renames.
- A new Package Archive reader is created for each `GetStream` call...
- If the package is deleted / renamed or modified NPE will show a message box. And in case of modifications automatically reloads the package.

fixes #434

### DIFF
--- a/PackageExplorer/MainWindow.xaml.cs
+++ b/PackageExplorer/MainWindow.xaml.cs
@@ -90,6 +90,8 @@ namespace PackageExplorer
         [Export]
         public IPackageEditorService EditorService { get; set; }
 
+        private string _tempFile;
+
         private bool HasUnsavedChanges
         {
             get
@@ -146,18 +148,21 @@ namespace PackageExplorer
         {
             IPackage package = null;
 
+            var tempFile = Path.GetTempFileName();
             try
             {
+                File.Copy(packagePath, tempFile, overwrite: true);
+
                 var extension = Path.GetExtension(packagePath);
                 if (extension.Equals(Constants.PackageExtension, StringComparison.OrdinalIgnoreCase))
                 {
-                    package = new ZipPackage(packagePath);
+                    package = new ZipPackage(tempFile);
                 }
                 else if (extension.Equals(Constants.ManifestExtension, StringComparison.OrdinalIgnoreCase))
                 {
-                    using (var str = ManifestUtility.ReadManifest(packagePath))
+                    using (var str = ManifestUtility.ReadManifest(tempFile))
                     {
-                        var builder = new PackageBuilder(str, Path.GetDirectoryName(packagePath));
+                        var builder = new PackageBuilder(str, Path.GetDirectoryName(tempFile));
                         package = builder.Build();
                     }
                 }
@@ -165,13 +170,22 @@ namespace PackageExplorer
                 if (package != null)
                 {
                     LoadPackage(package, packagePath, PackageType.LocalPackage);
+                    _tempFile = tempFile;
                     return true;
                 }
             }
             catch (Exception ex)
             {
+                package = null;
                 UIServices.Show(ex.Message, MessageLevel.Error);
                 return false;
+            }
+            finally
+            {
+                if (package == null && File.Exists(tempFile))
+                {
+                    File.Delete(tempFile);
+                }
             }
 
             return false;
@@ -244,6 +258,14 @@ namespace PackageExplorer
             {
                 currentViewModel.PropertyChanged -= OnPackageViewModelPropertyChanged;
                 currentViewModel.Dispose();
+            }
+            if (_tempFile != null)
+            {
+                if (File.Exists(_tempFile))
+                {
+                    File.Delete(_tempFile);
+                }
+                _tempFile = null;
             }
         }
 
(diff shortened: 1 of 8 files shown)


PackageExplorer/MainWindow.xaml.cs
```diff
@@ -235,6 +301,42 @@ private void OnPackageViewModelPropertyChanged(object sender, PropertyChangedEve
                     Content = RootLayout;
                 }
             }
+            else if (e.PropertyName == nameof(PackageViewModel.PackageSource))
+            {
```
campersau: `SaveAsXXX` will update the `PackageSource` property.

PackageExplorer/MainWindow.xaml.cs
```diff
@@ -235,6 +301,42 @@ private void OnPackageViewModelPropertyChanged(object sender, PropertyChangedEve
...
+                if (File.Exists(viewModel.PackageSource))
+                {
+                    InitPackageFileWatcher(viewModel.PackageSource);
+                }
+            }
+            else if (e.PropertyName == nameof(PackageViewModel.HasEdit))
+            {
+                if (viewModel.HasEdit && _packageFileStream == null && _packageFileWatcher != null)
```
campersau: Packages opened directly from package chooser are put into temp folder so they are not being watched and locked.

PackageViewModel/PackageViewModel.cs
```diff
@@ -110,6 +110,9 @@ internal ISettingsManager SettingsManager
             get { return _settingsManager; }
         }
 
+        public event EventHandler BeforeSave;
```
campersau: Required to ignore file watcher changes during save... Or is there an other way to do it?

PackageExplorer/MainWindow.xaml.cs
```diff
@@ -177,6 +184,63 @@ private bool OpenLocalPackageCore(string packagePath)
...
+        {
+            if (!_packageFileChanged) return;
+            _packageFileChanged = false;
+
+            if (_packageFileWatcher != null)
+            {
+                var packagePath = Path.Combine(_packageFileWatcher.Path, _packageFileWatcher.Filter);
+                if (File.Exists(packagePath))
+                {
+                    UIServices.Show("Package file has been changed", MessageLevel.Information);
```
campersau: Will put the strings into resources. Are they ok?

onovotny: What is the purpose of notifying the user it has been changed? Instead of a popup, can we alter the title or status bars?

PackageViewModel/PackageViewModel.cs
onovotny: What is the purpose of using a file watcher here?

campersau: To detect package file renames / deletes / changes so we can either dispose or reload the view model.

PackageExplorer/MainWindow.xaml.cs
campersau: Just to notify the user but the view model can be disposed or reloaded without a message.
What do you mean with updating the title? The view model is in an invalid state if the file is renamed / deleted / modified.

onovotny: what if we copy the file to a temp directory upon open, so it can never be renamed. We'd preserve the original path for save operations.

campersau: Ah yes, that would also work, but I wanted to do it the hard way 😄 
Will redo it.

campersau:
Looks like using a temp file is also solving #376 so we could revert #459

onovotny:
I think disposing of the streams is generally a good idea, so I don't see any reason to revert unless it's causing issues?

campersau:
Are you referring to #458 ? I was taking about #459 .

onovotny:
Oh, yeah, oops. Yeah, if this fixes that, then we can revert.

onovotny:
Is this ready for review/merging? Did you want to revert #459 in this PR?

campersau:
Reverted #459 and ready for review / merging.

PackageExplorer/MainWindow.xaml.cs
```diff
@@ -146,33 +148,45 @@ private bool OpenLocalPackageCore(string packagePath)
...
             {
+                package = null;
                 UIServices.Show(ex.Message, MessageLevel.Error);
                 return false;
             }
+            finally
+            {
+                if (package == null && File.Exists(tempFile))
+                {
+                    File.Delete(tempFile);
```
onovotny: This could still fail as File.Exists is a race condition, and File.Delete could throw. We should try/catch it and swallow.

PackageExplorer/MainWindow.xaml.cs
```diff
@@ -245,6 +259,14 @@ private void DisposeViewModel()
                 currentViewModel.PropertyChanged -= OnPackageViewModelPropertyChanged;
                 currentViewModel.Dispose();
             }
+            if (_tempFile != null)
+            {
+                if (File.Exists(_tempFile))
+                {
+                    File.Delete(_tempFile);
```
onovotny: Same as above.

PR #465, opened 2018-09-06
add support for settings in microsoft store package

### LINKED ISSUE #464: settings lost after update from microsoft store
**Type**: Bug

**NPE version**: 4.4.72

In case of a BUG:

- What is the current result?
After update from Microsoft Store the settings such as `MruPackageSources` are lost.

After update from Microsoft Store the settngs should not be lost.

---

The new way to set settings in desktop bridge applications is via [`ApplicationData`](https://docs.microsoft.com/en-us/windows/uwp/design/app-settings/store-and-retrieve-app-data) but it does not work in normal applications.

Also it looks like `Microsoft.DotNet.Framework.NativeImageCompiler` does not work if `Windows.winmd` is referenced.

See also: https://stackoverflow.com/questions/49561940/windows-forms-to-uwp-app-via-desktop-bridge-user-settings-are-lost-reset-on-uw

campersau:
fixes #464


PackageExplorer/App.xaml.cs
```diff
@@ -87,17 +88,6 @@ Task<IEnumerable<ICredentialProvider>> getProviders()
             
         }
 
-        private static void MigrateSettings()
```
campersau: I accidentally delete it.
Should I add it back?

onovotny: Is it needed? :)

campersau:
I am not sure I understand your second point.
I just use `ApplicationData.Current.LocalSettings` and `Settings.Default` and don't need a folder path.
Or do you mean to use the `NativeMethods.IsRunningAsUwp` here?
https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/blob/2a2e713b8b81458543132ab1dd017115cd9ad68d/Core/Repositories/MachineCache.cs#L138

PackageExplorer/App.xaml.cs
campersau: Yes, added back.

onovotny:
I don't think we need the native methods. We can try/catch getting the `ApplicationData.Current.` methods once and fall back to the existing mechanism?

PR #472, opened 2018-09-22
Port to .NET Core 3

onovotny:
Port NPE to run on .NET Core 3

Done, just waiting for a public preview build.


PackageExplorer/MainWindow.xaml.cs
```diff
@@ -464,7 +464,9 @@ private void OpenExternalLink(object sender, ExecutedRoutedEventArgs e)
             if (e.Parameter is X509Certificate2 cert)
             {
                 var hwnd = new WindowInteropHelper(this).Handle;
-                X509Certificate2UI.DisplayCertificate(cert, hwnd);
+                
+                // TODO: Replace when fixed
+                //X509Certificate2UI.DisplayCertificate(cert, hwnd);
```
campersau: It looks like `X509Certificate2UI` is in `System.Windows.Extensions` which is currently in preview in the `https://dotnet.myget.org/F/dotnet-core/api/v3/index.json` feed.

PR #480, opened 2018-10-15
Add Packaging support

rido-min:
This is a temp workaround to package a self-contained exe as APPX/MSIX.


onovotny:
Closing in favor of the other workaround

Issue #478, opened 2018-10-09
Show nuspec XML for signed packages


onovotny:
Signed packages hide the toolbar that enables the XML mode. While we don't want to enable changes, we should allow the XML to be viewed.

PR #485, opened 2018-10-18
Show nuspec XML for signed packages

samneirinck:
Adds the ability to view nuspec xml's for signed packages. I've made sure to disable the save commands where I could find them.

Not sure if this is sufficient, please let me know if I missed something.

Addresses #478

### DIFF
--- a/PackageExplorer/FileEditor.xaml
+++ b/PackageExplorer/FileEditor.xaml
@@ -82,9 +82,9 @@
                 </Button>
             
                 <!-- Save -->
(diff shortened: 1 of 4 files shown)


onovotny:
Hi @samneirinck Thanks for this! Two small things

- can you please sign the CLA so we can accept this?
- can you please hide the metadata button editor in this scenario? 
image

samneirinck:
Added some logic to hide the metadata button. I'm not convinced of the name, so feel free to change :)

Also signed the CLA in the meantime.

Thanks for the quick feedback!

onovotny:
Thanks!

PR #590, opened 2018-12-11
 Display icons for files too

### LINKED ISSUE #591: Display icons for files too
It would be nice to also display icons for the file items in the tree view, I've created [PR#590](https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/pull/590)

snipaste_2018-12-11_22-50-56

wangfu91:
It would be nice to also display icons for the file items in tree view.

snipaste_2018-12-11_22-50-56

Fixes #591

### DIFF
--- a/PackageViewModel/Utilities/FileHelper.cs
+++ b/PackageViewModel/Utilities/FileHelper.cs
@@ -177,7 +177,7 @@ namespace PackageExplorerViewModel
         private enum SHGFI
         {
             /// <summary>
-            /// get icon
+            /// Get icon
             /// </summary>
             Icon = 0x000000100,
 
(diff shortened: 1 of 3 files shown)


PackageViewModel/Utilities/FileHelper.cs
```diff
@@ -169,5 +171,157 @@ public static bool IsAssembly(string path)
...
+            /// <summary>
+            /// The handle that identifies a directory.
+            /// </summary>
+            FILE_ATTRIBUTE_DIRECTORY = 0x10,
+        }
+
+        /// <summary>
+        /// Maximal Length of unmanaged Windows-Path-strings
+        /// </summary>
+        private const int MAX_PATH = 260;
```
onovotny: This isn't really true anymore, FYI.

onovotny:
Thank you! I'll test this out and get it merged shortly.

PackageViewModel/Utilities/FileHelper.cs
wangfu91: Thanks, so the ```SizeConst = MAX_PATH``` should be removed from attribute?

onovotny: I'm not completely sure what the right answer is, I'll look into it.

vcsjones: The Win32 API sure does look like it is documented to be limited to `MAX_PATH`. https://docs.microsoft.com/en-us/windows/desktop/api/shellapi/ns-shellapi-_shfileinfow

Not sure how this API was updated to account for new limit, but you can't just remove it or change it because it will affect the size of the struct.

PR #596, opened 2018-12-14
Added repository fields to nuspec editor

### LINKED ISSUE #476: Add repository field to nuspec editor
Feature request 

 - Why do we need it?
It is a valid field already specifed in documentation https://docs.microsoft.com/en-us/nuget/reference/nuspec

 - Please provide an use case.
Adding this data enables use of SourceLink in vs2017 while debugging

vandersmissenc:
Added repository fields for type, url, branch, and commit to the nuspec editor page.

resolves #476

### DIFF
--- a/PackageExplorer/PackageMetadataEditor.xaml
+++ b/PackageExplorer/PackageMetadataEditor.xaml
@@ -84,7 +84,11 @@
                 <RowDefinition Height="Auto" />
                 <RowDefinition Height="Auto" />
                 <RowDefinition Height="Auto" />
-            </Grid.RowDefinitions>
+                <RowDefinition Height="Auto" />
+                <RowDefinition Height="Auto" />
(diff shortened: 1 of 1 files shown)


vandersmissenc:
Look like the chocolatey failed to generate a package due to the version string
`2018-12-14T14:04:34.5985210Z '5.0.121-5c042693e2' is not a valid version string.`

vandersmissenc:
It seems like the version string being generated doesn't meet the SemVer1 requirements outlined [here](https://github.com/semver/semver/blob/v1.0.0/semver.md) because the third segment contains alpha characters.

The most recent failed version string was:
```2018-12-14T14:34:37.0242801Z '5.0.12249c8cf0c4f' is not a valid version string.```

onovotny:
Yeah, it’s a bug in the latest NB.GV. Will pin to an earlier version until it’s fixed.

PR #638, opened 2019-02-08
Update FileContentInfo.cs

campersau:
fix using statement
introduced in https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/pull/637/files#diff-b6012059e9949534c39f49d35a72c703R1


onovotny:
thanks

PR #639, opened 2019-02-08
More fixes

onovotny:


PackageExplorer/app.config
```diff
@@ -7,7 +7,7 @@
   </configSections>
   <appSettings>
     <add key="bugsnag:apiKey" value="__BugsnagApiKey__" />
-    <add key="bugsnag:sourceRoots" value="__BugsnagSourceRoots__" />
+    <add key="bugsnag:sourceRoots" value="__BugsnagSourceRoots__;C:\dev\NuGet.Client\" />
```
campersau: Was is intentional?

onovotny: Yes. I build a local copy of NuGet which adds signing support for netcoreapp2.1. That's the path it uses.

PR #649, opened 2019-02-15
Revert "Search while typing."

onovotny:
Reverts NuGetPackageExplorer/NuGetPackageExplorer#619.

Typing to search and hitting enter doesn't always work fluently with this.


onovotny:
@wangfu91 I need to revert this PR for now since I think it's not working well with the search experience. The problem is that typing/hitting enter seems to conflict with this and it's not always picking up the correct search.

wangfu91:
@onovotny  I see, sorry about that, I did some tests before submitting the pr, but clearly not did enough tests, I will find a reliable way to implement this.

Issue #650, opened 2019-02-15
System.IO.IOException in PackageViewModel\PackageViewModel.cs:1224


onovotny:
## Error in NuGet Package Explorer

**System.IO.IOException** in **PackageViewModel\PackageViewModel.cs:1224**
The filename, directory name, or volume label syntax is incorrect. : 'C:\WINDOWS\system32\https:\api.nuget.org\v3\index.json'

[View on Bugsnag](https://app.bugsnag.com/oren-novotny-llc/nuget-package-explorer/errors/5c5ed7c275f6e2001a231150?event_id=5c66542100358dba02ab0000&i=gh&m=ci)

## Stacktrace

    PackageViewModel\PackageViewModel.cs:1224 - PackageExplorerViewModel.PackageViewModel.GetCurrentPackageTempFile()

[View full stacktrace](https://app.bugsnag.com/oren-novotny-llc/nuget-package-explorer/errors/5c5ed7c275f6e2001a231150?event_id=5c66542100358dba02ab0000&i=gh&m=ci)

*Created by Oren Novotny via Bugsnag*

onovotny:
@campersau was this related to one of the recent changes?

Seems like it started recently, but I haven't looked too closely yet.

PR #651, opened 2019-02-16
use separate variables for package source and package path

campersau:
this is a quick fix for #650 
Now the `PackageSource` in the `PackageViewModel` is always the package path. Which means that if a package is opened from a feed, the feed name is no longer shown in the package view but instead the temp path to the package.
If we want to preserve the previous behavior we also have to use separate variables for package source and package path in the `PackageViewModel` which would require more changes.

### DIFF
--- a/PackageViewModel/PackageViewModel.cs
+++ b/PackageViewModel/PackageViewModel.cs
@@ -46,6 +46,7 @@ namespace PackageExplorerViewModel
         private bool _isInEditMode;
         private RelayCommand<object>? _openContentFileCommand;
         private ICommand? _openWithContentFileCommand;
+        private string _packagePath;
         private string _packageSource;
         private RelayCommand? _publishCommand;
         private RelayCommand<object>? _renameContentCommand;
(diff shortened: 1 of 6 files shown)


onovotny:
I think we should use separate variables for the package source and path in the view model too so we hide the underlying details from the UI.

PackageViewModel/PackageViewModel.cs
```diff
@@ -83,7 +85,8 @@ public sealed class PackageViewModel : ViewModelBase, IDisposable
 
             _packageMetadata = new EditablePackageMetadata(_package, UIServices);
 
-            _packageSource = source;
```
campersau: @onovotny I noticed that you recently changed this to set the field instead of the property. Was this intended? By setting the property a `FileSystemWatcher` would have been created to watch changes to the file.

onovotny: That was to make a C# 8 warning go away. I didn’t realize the property did more, it should set the property then.

campersau: Ah, yeah I currently also get the warning. Should I just make the fields nullable then?

onovotny: No, it can be suppressed either with the ! post-fix operator or a pragma.

PR #739, opened 2019-07-30
Make text selectable in AssemblyMetadata panel

304NotModified:
Makes the text selectable in the "AssemblyMetadata" panel:

image

This is maybe a poor-mans solution, but haven't found another solution.

needed because: I need sometimes the PublicKeyToken and I don't like typing hex ;)

### DIFF
--- a/PackageExplorer/App.xaml
+++ b/PackageExplorer/App.xaml
@@ -90,6 +90,21 @@
                 </Style.Triggers>
             </Style>
 
+            <!-- Clone of SelectableTextBlockLikeStyle, but without triggers -->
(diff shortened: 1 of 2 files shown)


campersau:
We have also used the `SelectableTextBlockLikeStyle` style elsewhere.
Maybe you want to use it too. See https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/pull/617

304NotModified:
it's indeed almost the approach :) Do you think I should use the style in the code? Not sure if that make things more clear.

onovotny:
I like the idea of making the text selectable. If the style works, it would seem to be simpler than adding a new control for this?

304NotModified:
OK, I moved to style. Unfortunate I could not reuse SelectableTextBlockLikeStyle, as the triggers made the styling buggy (moving pixels when hovering)

PR #740, opened 2019-07-30
Don't show strange publish and wrong download count values for unlisted packages

304NotModified:
For unlisted packages the download count is wrong and the publish date is strange (i'm pretty sure I didn't released that day ;))

Download counts on nuget.org:

image

Solution for now: don't show those if unlisted.

image

I'm not sure if this is only an issue of nuget.org. Not sure if we need to test it and how. (another option is to test the date, e.g. before 1901, but not sure how to check the downloads then)

### DIFF
--- a/Core/Packages/PackageInfo.cs
+++ b/Core/Packages/PackageInfo.cs
@@ -20,7 +20,7 @@ namespace NuGetPe
         public string? Description { get; set; }
         public string? Summary { get; set; }
         public string? Authors { get; set; }
-        public int DownloadCount { get; set; }
+        public int? DownloadCount { get; set; }
         public DateTimeOffset? Published { get; set; }
 
         public string? IconUrl { get; set; }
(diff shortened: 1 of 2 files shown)


onovotny:
I think this is a v3 feed issue. There’s some way to get the data, as the site displays it, but I’m not sure exactly how yet. 

Makes sense to hide bad data until we can though. We should create a new issue to track displaying it correctly.

304NotModified:
OK, so this PR isn't a good solution I guess?
Then proposal:

- don't show date if before <1901
- don't show count if 0. It's already highly unlikely that a package is never downloaded ever ;)

304NotModified:
@onovotny 

Updated the logic. it's now:

- only use date if it's after 1900
- don't show downloadcount if: it unlisted, the versioninfo is null and the total is 0.

304NotModified:
@onovotny I'm doubting if we should clear the download count = 0. I see it's in more places wrong:

image

What do you think? 
(so proposal: if donwload = 0, then empty text in all cases)

onovotny:
That's fine.... I just am curious to know how the website gets the data. Must be a different feed api?

304NotModified:
OK it's here -> #746 :)

PR #768, opened 2019-08-07
Work around issue in .NET Core SDK 3.0 Preview 8

nguerrera:
See https://github.com/dotnet/sdk/issues/3512

This will be fixed in Preview 9

@onovotny Sorry for the trouble here.


onovotny:
@nguerrera Does this hurt on Preview 7, or will we need this just when we update to preview 8?

nguerrera:
> @nguerrera Does this hurt on Preview 7, or will we need this just when we update to preview 8?

@onovotny It should build find with preview 7, 8, or 9, but only be required for 8. The condition on the target makes it do nothing when the SDK does not have the issue.

It would be helpful to merge this since we have some folks that are doing testing against this repo, and they will be testing a mix of builds with and without the issue, looking for other issues.

PR #801, opened 2019-08-10
Added local packages search for static feeds

### LINKED ISSUE #607: No working Search for static feeds

**Type** (choose one):

- Bug

**NPE version**: (e.g. 5.x) (see Help -> about)

5.0.132+f33e9c233d

**OS version**: (Windows 7, 10 - 1803/17134, etc)

Windows 10

***Repro**
Open feed https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json.

Enter `Arcade` into search box and hit enter.

After a while the package list refreshes but contains all packages, not just those that include "Arcade".

304NotModified:
Including prelease filter, id filter, paging and ranked sorting. Fixes #607

Tested with https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json

Initial load is still a bit slow (~16 sec), but after that search and paging is fast!

(yes, this won't work with static feeds if they have <= 15 packages)

### DIFF
--- a/PackageViewModel/PackageChooser/PackageChooserViewModel.cs
+++ b/PackageViewModel/PackageChooser/PackageChooserViewModel.cs
@@ -11,6 +11,7 @@ using NuGet.Protocol;
 using NuGet.Protocol.Core.Types;
 using NuGetPackageExplorer.Types;
 using NuGetPe;
+using PackageExplorerViewModel.PackageSearch;
 
 namespace PackageExplorerViewModel
 {
@@ -238,7 +239,8 @@ namespace PackageExplorerViewModel
         public event EventHandler OpenPackageRequested = delegate { };
         public event EventHandler PackageDownloadRequested = delegate { };
 
-
+        private readonly PackageListCache<IPackageSearchMetadata> _packageListCache = new PackageListCache<IPackageSearchMetadata>();
+        
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Maintainability", "CA1506:AvoidExcessiveClassCoupling"),
          System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes")]
         private async Task LoadPackages()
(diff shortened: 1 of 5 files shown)


PackageViewModel/PackageSearch/LocalPackageSearcher.cs
```diff
@@ -0,0 +1,108 @@
...
+                    .OrderByDescending(it => it.score)
+                    .Select(it => it.package);
+            }
+
+            return packages;
+        }
+
+        private static int CalcScore(T package, string searchText)
+        {
+            var score = ScoreForEquals(package.Identity.Id, searchText);
```
304NotModified: FYI This isn't fully following https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/issues/607#issuecomment-517950958

PackageViewModel/PackageSearch/PackageListCache.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+using System.Collections.Generic;
+using NuGet.Protocol.Core.Types;
+
+namespace PackageExplorerViewModel.PackageSearch
+{
+    internal class PackageListCache<T>  where T : IPackageSearchMetadata
```
304NotModified: Better cache proposals welcome :)

PackageViewModel/PackageSearch/SearchContext.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System;
+using NuGet.Protocol.Core.Types;
+
+namespace PackageExplorerViewModel.PackageSearch
+{
+    public class SearchContext
```
304NotModified: FYI: Wrapper for current search context for easy passing deeper

PackageViewModel/PackageChooser/PackageChooserViewModel.cs
```diff
@@ -370,15 +372,6 @@ private void ShowPackages(IEnumerable<IPackageSearchMetadata> packages, int pack
             if (ActiveRepository != null)
             {
                 var ar = ActiveRepository;
-                var skip = beginPackage - 1;
```
304NotModified: FYI, not needed anymore. This is done in another (and better) place

PackageViewModel/PackageSearch/LocalPackageSearcher.cs
onovotny: What are the differences?

onovotny: Can the package scoring be parellelized to speed it up? Parallel.ForEach, etc?

304NotModified: Maybe, but it's now already blazing fast. (instant experience with 1900 packages). The slow part is already done (download JSON  + parse all JSON).

304NotModified:
When we are there, please squashed PR merge :)

PR #811, opened 2019-08-12
Infinite scrolling in PackageChooser

### LINKED ISSUE #413: Infinite Scrolling for Package Chooser Dialog
The Package Chooser dialog should support infinite scrolling instead of needing an explicit pager.

The code can be adapted from:
https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml
https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs
https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollListBox.cs

campersau:
Very basic implementation using the existing UI.

Known issues:
- ~If the PackageChooser window is higher than the height of the initial 15 packages, no more packages are loaded automatically. You need to resize the window to make the scrollbar visible and scroll to the end.~

fixes #413


onovotny:
Sweet!

What's needed to fix that issue? It would seem to bite people accidentally?

PackageExplorer/PackageChooser/PackageChooserDialog.xaml.cs
```diff
@@ -91,11 +91,7 @@ private void InvokeSearch(string? searchTerm)
 
         private async void Window_Loaded(object sender, RoutedEventArgs e)
         {
-            if (string.IsNullOrEmpty(_pendingSearch))
-            {
-                await Dispatcher.BeginInvoke(new Action(LoadPackages), DispatcherPriority.Background);
```
campersau: This was a bug because the `PackageSourceBox_OnSelectionChanged` handler is also always called every time the window is shown which caused the packages to load twice (discarding the first load).

campersau:
I have fixed the issue with not automatically loading more packages if scrollbar is not visible.

304NotModified:
cool! works nice! Maybe a small details would be the style/position loader when loading a new "page".

image

Tested also with the static feed, works very nice (no loaders there after first "load")

campersau:
@304NotModified Yes, the loading UI on top of the packages isn't nice. Suggestions are welcome (I am not good at UI design 😄 ). But it can also be done in another PR.

304NotModified:
> But it can also be done in another PR.

Yes that sound like a good plan to me. It's a really small detail

PackageExplorer/PackageChooser/PackageChooserDialog.xaml.cs
```diff
@@ -220,5 +210,27 @@ private void PackageSourceBox_OnSelectionChanged(object sender, SelectionChanged
...
+
+        private void ListBoxPackages_ScrollChanged(object sender, ScrollChangedEventArgs e)
+        {
+            if (e.VerticalChange != 0 && e.OriginalSource is ScrollViewer scrollViewer && scrollViewer.VerticalOffset == scrollViewer.ScrollableHeight)
+            {
+                if (scrollViewer.VerticalOffset == 0 && scrollViewer.ScrollableHeight == 0) // scroll back to top if packages got cleared
+                {
+                    scrollViewer.ScrollToTop();
+                }
+                else if (_viewModel.IsEditable && _viewModel.HasMore)
```
campersau: Maybe move this check to the LoadedCommand CanExecute action.
And rename it to `LoadMoreCommand`.

304NotModified: 👍

onovotny:
I just merged this so there can be a new PR for the loading progress bar.

onovotny:
Does anyone have time to submit an update to the package load progress (perhaps to a spinner somewhere?) so we can get a new release out that's a bit more polished?

304NotModified:
Not me.  I'm  busy with NLog the next few days.

304NotModified:
IMO it could be released without it. Pretty awesome features are added in this release :)

campersau:
I am back from vacation and can work on it. Will try to make it similar to the VS package browser.

PR #818, opened 2019-08-22
Refactor package chooser loading

campersau:
Adding a fake item which shows the loading indicator inside the list instead of on top of it.

image

image

I can also use the [loading spinner](https://github.com/NuGet/NuGet.Client/blob/eab7f0366f74f30b7ceb9b23f63f1f23adfc45f9/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/Spinner.xaml) from VS instead of the progressbar.


onovotny:
I do like VS's loading spinner

campersau:
Updated to use VS package manager spinner.

campersau:
@onovotny yes, it is.

PackageExplorer/Controls/Spinner.xaml.cs
```diff
@@ -14,7 +14,9 @@ internal class EllipseDetails
         public double Height { get; set; }
         public double Left { get; set; }
         public double Top { get; set; }
+#pragma warning disable CS8618 // Non-nullable field is uninitialized. Consider declaring as nullable.
         public Brush Fill { get; set; }
+#pragma warning restore CS8618 // Non-nullable field is uninitialized. Consider declaring as nullable.
```
304NotModified: @onovotny do you see warnings that others aren't seeing?

304NotModified: It's configured as warning (and not error), isn't?

PR #826, opened 2019-09-23
Update to stable version of .net core 3

onovotny:


304NotModified:
Related https://developercommunity.visualstudio.com/content/problem/745907/msbuild-163-on-hosted-azure-build-pipeline.html

PR #841, opened 2019-10-04
Don't use SocketHttpHandler because of authentification issues with …

campersau:
…GitHub NuGet feeds

The `SocketHttpHandler` has some issues with resending the same request multiple times. Which is what `NuGet.Client` does for authentification. Because of that the authentification for GitHub NuGet package feeds does not work. Because we are using the default credentials for the first authorization request and after that the `SocketHttpClient` does not send the updated credentials from the user. Because the GitHub NuGet API does not send a `Www-Authenticate` header if an `Authorization` header was provided.

https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/blob/e95bd234fdaa74e2367a29dc141ec8275e2d08d0/PackageExplorer/MefServices/CredentialManager.cs#L45

To try it out use your GitHub NuGet feed (`https://nuget.pkg.github.com/GITHUBUSER/index.json`).

See https://github.com/dotnet/corefx/issues/25163#issuecomment-538009706 for more details

### DIFF
--- a/PackageExplorer/App.xaml.cs
+++ b/PackageExplorer/App.xaml.cs
@@ -25,6 +25,10 @@ namespace PackageExplorer
         public App()
 #pragma warning restore CS8618 // Non-nullable field is uninitialized.
         {
+            // Don't use the SocketHttphandler because it has some authentification issues accessing feeds like GitHub NuGet
+            // see https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/pull/841 for more details
+            AppContext.SetSwitch("System.Net.Http.UseSocketsHttpHandler", false);
+
             DiagnosticsClient.Initialize();
         }
 


PackageExplorer/App.xaml.cs
```diff
@@ -25,6 +25,7 @@ public partial class App : Application
         public App()
 #pragma warning restore CS8618 // Non-nullable field is uninitialized.
         {
+            AppContext.SetSwitch("System.Net.Http.UseSocketsHttpHandler", false);
```
304NotModified: I prefer to have a comment in code for this decision. Otherwise the reason could be hard to find in the future.

304NotModified: Thanks!

onovotny:
Does this need to be here or in the underlying NuGet libraries?

campersau:
Since this approche uses static switches it is better to add it in each application.
But there are other solutions as well:

- corefx fixes the issue linked above (would be the best)
- rewrite NuGet.Client http handlers to not reuse the same request every time by e.g. cloning it.
The [retry handler already creates new requests each time](https://github.com/NuGet/NuGet.Client/blob/caac28293a4032de8629b39597a179df988a612a/src/NuGet.Core/NuGet.Protocol/HttpSource/HttpRetryHandler.cs#L47) but not the [authentification](https://github.com/NuGet/NuGet.Client/blob/caac28293a4032de8629b39597a179df988a612a/src/NuGet.Core/NuGet.Protocol/HttpSource/HttpSourceAuthenticationHandler.cs), [proxy authentification](https://github.com/NuGet/NuGet.Client/blob/caac28293a4032de8629b39597a179df988a612a/src/NuGet.Core/NuGet.Protocol/HttpSource/ProxyAuthenticationHandler.cs) and [STS](https://github.com/NuGet/NuGet.Client/blob/caac28293a4032de8629b39597a179df988a612a/src/NuGet.Core/NuGet.Protocol/HttpSource/StsAuthenticationHandler.cs) handlers (and maybe more).

onovotny:
@campersau Might want to file an issue on the NuGet repo so they're aware and can track it?

campersau:
See https://github.com/NuGet/Home/issues/8661

PR #846, opened 2019-10-16
Code analysis

onovotny:
Add FxCop Analyzers and related fixes

### DIFF
--- a/Core/AssemblyMetadata/AssemblyMetadataParser.cs
+++ b/Core/AssemblyMetadata/AssemblyMetadataParser.cs
@@ -241,7 +241,7 @@ namespace NuGetPe.AssemblyMetadata
                     }
                 }
 
-                throw new UnknownTypeException();
+                throw new UnknownTypeException(@"Type {type} is of uknown TypeCode");
             }
         }
 
@@ -252,7 +252,17 @@ namespace NuGetPe.AssemblyMetadata
 
         private class UnknownTypeException : InvalidOperationException
         {
+            public UnknownTypeException(string message) : base(message)
+            {
+            }
 
+            public UnknownTypeException(string message, Exception innerException) : base(message, innerException)
+            {
+            }
+
+            public UnknownTypeException()
+            {
+            }
         }
     }
 }
--- a/Core/AssemblyMetadata/AssemblyMetadataParser.cs
+++ b/Core/AssemblyMetadata/AssemblyMetadataParser.cs
@@ -241,7 +241,7 @@ namespace NuGetPe.AssemblyMetadata
                     }
                 }
 
-                throw new UnknownTypeException(@"Type {type} is of uknown TypeCode");
+                throw new UnknownTypeException($"Type '{type}' is of unknown TypeCode");
             }
         }
 
--- a/Core/Extensions/CollectionExtensions.cs
+++ b/Core/Extensions/CollectionExtensions.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Linq;
 using NuGet.Packaging;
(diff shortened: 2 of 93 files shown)


Core/AssemblyMetadata/AssemblyMetadataParser.cs
```diff
@@ -241,7 +241,7 @@ public PrimitiveTypeCode GetUnderlyingEnumType(string type)
                     }
                 }
 
-                throw new UnknownTypeException();
+                throw new UnknownTypeException(@"Type {type} is of uknown TypeCode");
```
304NotModified: Small typo here ;)

304NotModified: Maybe pass the Type as parameter to the exception?

Core/Extensions/CollectionExtensions.cs
```diff
@@ -9,12 +9,22 @@ public static class CollectionExtensions
     {
         public static void CopyTo<T>(this IEnumerable<T> sourceCollection, ICollection<T> targetCollection)
         {
+            if (sourceCollection is null)
```
304NotModified: Do you really prefer `is` over `==` here?

PackageExplorer/Converters/UpperCaseConverter.cs
```diff
@@ -8,10 +8,10 @@ public class UppercaseConverter : IValueConverter
     {
         #region IValueConverter Members
 
-        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+        public object? Convert(object value, Type targetType, object parameter, CultureInfo culture)
         {
             var s = (string)value;
-            return s.ToUpper(culture);
+            return s?.ToUpper(culture);
```
304NotModified: Could this be really null in c# 8?

PackageExplorer/PackageViewer.xaml.cs
```diff
@@ -626,6 +625,7 @@ private bool HandleDataObject(PackageFolder? folder, IDataObject data, bool copy
             return false;
         }
 
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Reliability", "CA2000:Dispose objects before losing scope", Justification = "<Pending>")]
```
304NotModified: What's "pending" means here?

Core/Extensions/CollectionExtensions.cs
304NotModified: It's now inconsistent in the code, something there is a `is null` and `== null` check in the same method.

304NotModified:
Nice changes! Spotted some small interesting changes, but looks great overall :)

Core/AssemblyMetadata/AssemblyMetadataParser.cs
campersau: typo**s** ;)

```suggestion
                throw new UnknownTypeException($"Type {type} is of unknown TypeCode");
```

Core/Extensions/CollectionExtensions.cs
onovotny: `is` is generated by the quick fix and it's not wrong... so seemed easiest to use it.

PackageExplorer/Converters/UpperCaseConverter.cs
onovotny: Nullables are currently analysis only. Nothing enforces it, so that's why the null checks are needed.

PackageExplorer/PackageViewer.xaml.cs
onovotny: that's the default comment put in by the suppression gesture.

PR #847, opened 2019-10-17
Feature/native targets filename rule

### LINKED ISSUE #845: Check that the targets file have the same name as package id for native NuGet packages
**Type**: Bug or Feature request

**NPE version**: NuGet Package Explorer - Store - X64 - (5.4.1+e95bd234fd)

**OS version**: Windows 10 Pro (10.0.18362)

I`m not sure if this is a bug or feature request.

I was publishing native NuGet package: https://www.nuget.org/packages/Platform.Collections.Methods.TemplateLibrary

And I have set a name for `targets` file in the `build\native` directory to `PlatformCollectionsMethods` and `Platform.Collections.Methods` but not `Platform.Collections.Methods.TemplateLibrary`.
This caused the issue of not including the path to library's header files in the vcxproj, where the package was included.

Before publishing of each version I have used `Analyze Package` menu item and got result of `0 issues found.`

So to save time from future mistakes, I suggest to include the check that ensures that `targets` file in `build\native` directory has has the same name as the package id in the `nuspec` file.

That will ensure that `targets` file will be successfully included into `vcxproj` when the package is used.

I had the problem in the versions 0.0.1-0.0.2 of [Platform.Collections.Methods.TemplateLibrary](https://www.nuget.org/packages/Platform.Collections.Methods.TemplateLibrary) and I fixed it in the 0.0.3 version.

aerotog:
This PR adds a new analysis rule that requires native Nuget packages to have their .targets file match their package name.

This resolves #845.

### DIFF
--- a/PackageViewModel/PackageAnalyzer/MisnamedNativeBuildFileRule.cs
+++ b/PackageViewModel/PackageAnalyzer/MisnamedNativeBuildFileRule.cs
@@ -1,0 +1,53 @@
+﻿using System;
+using System.Collections.Generic;
+using System.ComponentModel.Composition;
+using System.IO;
+using System.Linq;
+using NuGetPackageExplorer.Types;
+using NuGetPe;
+
+namespace PackageExplorerViewModel.Rules
+{
+    [Export(typeof(IPackageRule))]
+    internal class MisnamedNativeBuildFileRule : IPackageRule
+    {
+        #region IPackageRule Members
+
+        public IEnumerable<PackageIssue> Validate(IPackage package, string packagePath)
+        {
+            var files =
+                package.GetFiles().Where(x =>
+                    x.Path.EndsWith(".props", StringComparison.OrdinalIgnoreCase) ||
+                    x.Path.EndsWith(".targets", StringComparison.OrdinalIgnoreCase));
+
+            foreach (var file in files)
+            {
+                var path = file.Path;
+                var segments = path.Split('\\');
+
+                var frameworkFolder = segments[^2];
+                var filename = segments.Last();
+                var filenameWithoutExtension = Path.GetFileNameWithoutExtension(filename);
+
+                if (string.Equals(frameworkFolder, "native", StringComparison.OrdinalIgnoreCase) &&
+                    filenameWithoutExtension != package.Id)
+                {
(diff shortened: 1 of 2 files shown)


PackageViewModel/PackageAnalyzer/NativeTargetsFilenameRule.cs
```diff
@@ -0,0 +1,52 @@
...
+            foreach (var file in targetsFiles)
+            {
+                var path = file.Path;
+                var segments = path.Split('\\');
+
+                var frameworkFolder = segments[^2];
+                var filename = Path.GetFileNameWithoutExtension(path);
+
+                //if (path.EndsWith(".dll", StringComparison.OrdinalIgnoreCase) ||
+                //path.EndsWith(".winmd", StringComparison.OrdinalIgnoreCase))
```
onovotny: Please remove the commented out code

PackageViewModel/PackageAnalyzer/NativeTargetsFilenameRule.cs
```diff
@@ -0,0 +1,52 @@
...
+                if (string.Equals(frameworkFolder, "native", StringComparison.InvariantCultureIgnoreCase) &&
```
onovotny: This should probably be `OrdinalIgnoreCase` since `native` is a known folder.

PackageViewModel/PackageAnalyzer/NativeTargetsFilenameRule.cs
```diff
@@ -0,0 +1,52 @@
...
+{
+    [Export(typeof(IPackageRule))]
+    internal class NativeTargetsFilenameRule : IPackageRule
+    {
+        #region IPackageRule Members
+
+        public IEnumerable<PackageIssue> Validate(IPackage package, string packagePath)
+        {
+            var targetsFiles =
+                package.GetFiles().Where(x => x.Path.EndsWith(".targets", StringComparison.OrdinalIgnoreCase));
```
onovotny: Should this be `.props` as well?

aerotog: The rule now validates both .targets and .props.

aerotog:
PR updated and ready for review.

onovotny:
Thanks!

Issue #805, opened 2019-08-11
Multi-select in Package Contents list


jsimmonstx:
NPE version: 5.2.88
OS version: Win7 (in a VirftualBox VM, on a Linux host)
Installed from: file download

Feature Request:
Add the ability to select multiple files/folders in the Package Content list

Why do we need it?
Because we shouldn't be able to delete multiple files at once. Right now, it's one at a time, or delete the whole folder and start again. 

Please provide an use case:
A folder with more just a few files makes for a tedious editing session. There's no tangible reason for the single select restriction.

huffSamuel:
I use this tool and would like to contribute to it. I've been working on implementing this feature and have some questions about the expected behavior for the multiselect. Obviously not all functionality can apply to a group select and so far I have implemented the following:
- Drag/drop (move & copy)
- Copy/paste
- Delete

Are there any other behaviors that we should apply group behavior to?

mangelovic:
Hello,

is this issue still open for contribution (actual)?

onovotny:
Yes, please!

mangelovic:
![image](https://user-images.githubusercontent.com/56362720/66509527-8a409c80-ead3-11e9-8284-3ee6e5b9f045.png)

Just to be clear, in this case, multiselect should be available for deletion etc. right?

mangelovic:
Is it necessary to make just deletion for multi-select? Because maybe alsoCopy functionality makes sense to do also with that, but not sure if it is requested?

PR #851, opened 2019-11-02
Multiselect in package contents list

mangelovic:
-> usage of MultiSelectTreeView extension of basic TreeView element
-> overriden TreeView style for better selection of items (PackageViewer.xaml)
-> adjusted delete functionality for multi selection

Multi selection in Package Content List was implemented. I was thinking also to disable other options (allow just delete) in ContextMenu's, but I was unable to bind it well into model (if it will be requested, I can work it out). I focused on availability of multiselection for delete purposes, as was described in Issue (https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/issues/805).

### DIFF
--- a/PackageExplorer/Controls/MultiSelectTreeView.cs
+++ b/PackageExplorer/Controls/MultiSelectTreeView.cs
@@ -57,15 +57,20 @@ namespace PackageExplorer.Controls
             if (treeViewItem != null && treeView != null)
             {
                 var selectedItems = GetSelectedItems(treeView);
+
                 if (selectedItems != null)
                 {
                     if (GetIsItemSelected(treeViewItem))
                     {
-                        selectedItems.Add(treeViewItem.Header);
+                        if (selectedItems.Contains(treeViewItem.DataContext))
+                        {
+                            return;
+                        }
+                        selectedItems.Add(treeViewItem.DataContext);
                     }
                     else
                     {
-                        selectedItems.Remove(treeViewItem.Header);
+                        selectedItems.Remove(treeViewItem.DataContext);
                     }
                 }
             }
--- a/PackageExplorer/Controls/MultiSelectTreeView.cs
+++ b/PackageExplorer/Controls/MultiSelectTreeView.cs
@@ -132,7 +132,7 @@ namespace PackageExplorer.Controls
             var currentTime = DateTime.Now;
 
             // This is to prevent case, when holding CONTROL, GotFocus triggers twice (not all the time), so selected item do not get unselected right after
-            if ((currentTime - _lastTime).Milliseconds < 100)
+            if ((currentTime - _lastTime).Milliseconds < 5)
             {
                 return;
             }
--- a/PackageExplorer/Controls/MultiSelectTreeView.cs
+++ b/PackageExplorer/Controls/MultiSelectTreeView.cs
@@ -13,9 +13,19 @@ namespace PackageExplorer.Controls
     {
         public MultiSelectTreeView()
         {
-            GotFocus += OnTreeViewItemGotFocus;
-            PreviewMouseLeftButtonDown += OnTreeViewItemPreviewMouseDown;
-            PreviewMouseLeftButtonUp += OnTreeViewItemPreviewMouseUp;
+            Loaded += (s, a) =>
+            {
+                GotFocus += OnTreeViewItemGotFocus;
+                PreviewMouseLeftButtonDown += OnTreeViewItemPreviewMouseDown;
+                PreviewMouseLeftButtonUp += OnTreeViewItemPreviewMouseUp;
+            };
+
+            Unloaded += (s, a) =>
+            {
+                GotFocus -= OnTreeViewItemGotFocus;
+                PreviewMouseLeftButtonDown -= OnTreeViewItemPreviewMouseDown;
+                PreviewMouseLeftButtonUp -= OnTreeViewItemPreviewMouseUp;
+            };
         }
 
         private static TreeViewItem _selectTreeViewItemOnMouseUp;
--- a/PackageExplorer/Controls/MultiSelectTreeView.cs
+++ b/PackageExplorer/Controls/MultiSelectTreeView.cs
@@ -28,14 +28,19 @@ namespace PackageExplorer.Controls
             };
         }
 
-        private static TreeViewItem _selectTreeViewItemOnMouseUp;
-
+        private static TreeViewItem? _selectTreeViewItemOnMouseUp;
+        private static DateTime _lastTime = DateTime.Now;
 
         public static readonly DependencyProperty IsItemSelectedProperty = DependencyProperty.RegisterAttached("IsItemSelected", typeof(Boolean), typeof(MultiSelectTreeView), new PropertyMetadata(false, OnIsItemSelectedPropertyChanged));
 
-        public static bool GetIsItemSelected(TreeViewItem element)
+        public static bool GetIsItemSelected(TreeViewItem? element)
         {
-            return (bool)element.GetValue(IsItemSelectedProperty);
+            if (element != null)
+            {
+                return (bool)element.GetValue(IsItemSelectedProperty);
+            }
+
+            return false;
         }
 
         public static void SetIsItemSelected(TreeViewItem element, Boolean value)
@@ -70,12 +75,26 @@ namespace PackageExplorer.Controls
 
         public static IList GetSelectedItems(TreeView element)
         {
-            return (IList)element.GetValue(SelectedItemsProperty);
+            if (element != null)
+            {
+                return (IList)element.GetValue(SelectedItemsProperty);
+            }
+            else
+            {
+                throw new ArgumentException("TreeView element is null. Cannot set value for null element.");
+            }
         }
 
         public static void SetSelectedItems(TreeView element, IList value)
         {
-            element.SetValue(SelectedItemsProperty, value);
+            if (element != null)
+            {
+                element.SetValue(SelectedItemsProperty, value);
+            }
+            else
+            {
+                throw new ArgumentException("TreeView element is null. Cannot set value for null element.");
+            }
         }
 
         private static readonly DependencyProperty StartItemProperty = DependencyProperty.RegisterAttached("StartItem", typeof(TreeViewItem), typeof(MultiSelectTreeView));
@@ -86,7 +105,7 @@ namespace PackageExplorer.Controls
             return (TreeViewItem)element.GetValue(StartItemProperty);
         }
 
-        private static void SetStartItem(TreeView element, TreeViewItem value)
+        private static void SetStartItem(TreeView element, TreeViewItem? value)
         {
             element.SetValue(StartItemProperty, value);
         }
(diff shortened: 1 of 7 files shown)


PackageExplorer/Controls/MultiSelectTreeView.cs
```diff
@@ -0,0 +1,291 @@
...
+    public class MultiSelectTreeView : TreeView
+    {
+        public MultiSelectTreeView()
+        {
+            GotFocus += OnTreeViewItemGotFocus;
+            PreviewMouseLeftButtonDown += OnTreeViewItemPreviewMouseDown;
+            PreviewMouseLeftButtonUp += OnTreeViewItemPreviewMouseUp;
+        }
+
+        private static TreeViewItem _selectTreeViewItemOnMouseUp;
```
campersau: We are using the [new nullable reference types feature from C#](https://docs.microsoft.com/dotnet/csharp/nullable-references), so you need to declare this (and a lot of other variables in this file) as such.
If you are using the latest VS release you should see a lot of warnings regarding this.

Also, does it need to be `static`? Maybe you can cast the `sender`s in the events to a `MultiSelectTreeView` and set this instance field?

```suggestion
        private TreeViewItem? _selectTreeViewItemOnMouseUp;
```

PackageExplorer/Controls/MultiSelectTreeView.cs
```diff
@@ -0,0 +1,291 @@
...
+using System.Windows.Media;
+using System.Windows.Media.Media3D;
+
+namespace PackageExplorer.Controls
+{
+    public class MultiSelectTreeView : TreeView
+    {
+        public MultiSelectTreeView()
+        {
+            GotFocus += OnTreeViewItemGotFocus;
```
campersau: Maybe use this pattern to properly unsubscribe from the events?
https://stackoverflow.com/questions/25005355/pattern-to-unsubscribe-events-in-customcontrols

PackageViewModel/PackageViewModel.cs
```diff
@@ -601,9 +622,40 @@ private void DeleteContentExecute(object parameter)
         {
             DiagnosticsClient.TrackEvent("PackageViewModel_DeleteContentExecute");
 
-            if ((parameter ?? SelectedItem) is PackagePart file)
+            // Null objects are put sometimes into List, so to not count them
```
campersau: Do you know how to reproduce this?

mangelovic: As I wrote down into commit description. This occur only when user creates new file / folder into existing folder. If I create new file and not putting it right into some folder, it does not happen. I tried to check it out, or remove those ones on creating of new file, but it did not work.

PackageViewModel/PackageViewModel.cs
```diff
@@ -235,6 +239,21 @@ public ICollection<PackagePart> PackageParts
...
         }
 
+        public List<object?> SelectedItems
+        {
+            get { return _selectedItems; }
+            set
+            {
+                if (!_selectedItems.Contains(value))
+                {
+                    _selectedItems.Add(value);
```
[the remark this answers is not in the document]
mangelovic: Yes it was my mistake, as I just made "copy" of logic from SelectedItem and focused on functionality so...fixed now

PackageExplorer/Controls/MultiSelectTreeView.cs
mangelovic: Gonna try to remake these warnings, but when I tried to change few things, custom component functionality seemed to be a bit broken.

PackageExplorer/Controls/MultiSelectTreeView.cs
mangelovic: Used pattern which you recommended :)

PackageViewModel/PackageViewModel.cs
```diff
@@ -235,6 +239,21 @@ public ICollection<PackagePart> PackageParts
             }
         }
 
+        public List<object?> SelectedItems
+        {
+            get { return _selectedItems; }
+            set
+            {
+                if (!_selectedItems.Contains(value))
```
campersau: Sorry about not beeing clear enough the first time but this line also looks  a bit strange :)
And as mentioned before you might be able to just get rid of the setter because in my short testing it never got called.

mangelovic: Tested it out without having setter and works fine. My bad not seeing this. Fixed :)

PackageExplorer/Controls/MultiSelectTreeView.cs
```diff
@@ -0,0 +1,329 @@
...
+            var treeViewItem = d as TreeViewItem;
+            var treeView = FindTreeView(treeViewItem);
+            if (treeViewItem != null && treeView != null)
+            {
+                var selectedItems = GetSelectedItems(treeView);
+                if (selectedItems != null)
+                {
+                    if (GetIsItemSelected(treeViewItem))
+                    {
+                        selectedItems.Add(treeViewItem.Header);
```
campersau: I have debugged the `null` items and it looks like the problem is here.
Why are you using the `Header` instead of the `DataContext` here?
When I used the `DataContext` then no `null` items are added and everything works like before.

mangelovic: I have switched Header to DataContext. But it was adding duplicates of treeViewItem.DataContext into SelectedItems list. So additionally I added check, whether it is not adding same item twice. :)

onovotny:
@campersau Feel free to merge once you're satisfied.

PackageExplorer/Controls/MultiSelectTreeView.cs
```diff
@@ -0,0 +1,334 @@
...
+
+        public static void SetSelectedItems(TreeView element, IList value)
+        {
+            if (element != null)
+            {
+                element.SetValue(SelectedItemsProperty, value);
+            }
+            else
+            {
+                throw new ArgumentException("TreeView element is null. Cannot set value for null element.");
```
304NotModified: (minor) This should be an argumentNullException?

PackageExplorer/Controls/MultiSelectTreeView.cs
```diff
@@ -0,0 +1,334 @@
...
+            else if (treeViewItem != null)
+            {
+                for (int i = 0; i < treeViewItem.Items.Count; i++)
+                {
+                    var item = treeViewItem.ItemContainerGenerator.ContainerFromIndex(i) as TreeViewItem;
+                    if (item != null)
+                    {
+                        SetIsItemSelected(item, false);
+                        DeSelectAllItems(null, item);
+                    }
```
304NotModified: could we  easily deduplicate this by using a variable for `treeView` / `treeViewItem`? (loops are duplicated)

PackageExplorer/Controls/MultiSelectTreeView.cs
```diff
@@ -0,0 +1,334 @@
...
+                {
+                    var item = treeView.ItemContainerGenerator.ContainerFromIndex(i) as TreeViewItem;
+                    if (item != null)
+                    {
+                        allItems.Add(item);
+                        GetAllItems(null, item, allItems);
+                    }
+                }
+            }
+            else if (treeViewItem != null)
```
304NotModified: idem duplicate loops

PackageViewModel/PackageViewModel.cs
```diff
@@ -94,6 +96,8 @@ public sealed class PackageViewModel : ViewModelBase, IDisposable
             _isSigned = _packageMetadata.IsSigned;
 
             RootFolder = PathToTreeConverter.Convert(_package.GetFiles().ToList(), this);
+
+            //EnableOtherFunctions = false;
```
304NotModified: dead code :)

PackageExplorer/Controls/MultiSelectTreeView.cs
```diff
@@ -0,0 +1,333 @@
...
+            if (Mouse.LeftButton == MouseButtonState.Pressed && GetIsItemSelected(treeViewItem) && Keyboard.Modifiers != ModifierKeys.Control)
+            {
+                _selectTreeViewItemOnMouseUp = treeViewItem;
+                return;
+            }
+
+            var currentTime = DateTime.Now;
+
+            // This is to prevent case, when holding CONTROL, GotFocus triggers twice (not all the time), so selected item do not get unselected right after
+            if ((currentTime - _lastTime).Milliseconds < 100)
```
campersau: This also prevents keyboard navigation.
If you hold the up arrow key or down arrow key inside the tree view then the selection lags and get's out of sync if you press the other arrow key afterwards.

How exactly are you reproducing the CONTROL, GotFocus problem? For me it seems to work fine without it.

PackageExplorer/PackageViewer.xaml
```diff
@@ -580,6 +580,119 @@
...
+                                    <ColumnDefinition Width="Auto"/>
+                                    <ColumnDefinition Width="*"/>
+                                </Grid.ColumnDefinitions>
+                                <Grid.RowDefinitions>
+                                    <RowDefinition Height="Auto"/>
+                                    <RowDefinition/>
+                                </Grid.RowDefinitions>
+                                <ToggleButton x:Name="Expander" ClickMode="Press" IsChecked="{Binding IsExpanded, RelativeSource={RelativeSource TemplatedParent}}" Style="{StaticResource ExpandCollapseToggleStyle}"/>
+                                <Border x:Name="Bd" BorderBrush="{TemplateBinding BorderBrush}" BorderThickness="{TemplateBinding BorderThickness}" Background="{TemplateBinding Background}" Grid.Column="1" Padding="{TemplateBinding Padding}" SnapsToDevicePixels="true">
+                                    <ContentPresenter x:Name="PART_Header" ContentSource="Header" HorizontalAlignment="{TemplateBinding HorizontalContentAlignment}" SnapsToDevicePixels="{TemplateBinding SnapsToDevicePixels}"/>
```
campersau: Currently deselecting a folder while holding the CONTROL key does not work because the `OriginalSource` is a `Run` which is not supported by `VisualTreeHelper.GetParent`.
Adding `IsHitTestVisible="False"` here to worked for me.

PackageViewModel/PackageViewModel.cs
```diff
@@ -623,9 +630,37 @@ private void DeleteContentExecute(object parameter)
...
+                }
+
+                for (int i = SelectedItems.Count - 1; i >= 0; i--)
+                {
+                    if (SelectedItems[i] is PackagePart file)
+                    {
+                        file.Delete(false);
+                        continue;
+                    }
+                    if (SelectedItems[i] == null)
```
campersau: I think this can also be removed now.

PackageExplorer/Controls/MultiSelectTreeView.cs
mangelovic: Really could not find exactly why is this happening. 1st time you start app, it is okay without this line of code, 2nd time (it does not have to), but basically it gets triggered twice with CONTROL down on click, which results in not (un)selecting item you clicked on. I tried it with reduced number to 5 milliseconds, which worked for me still ok. Also I did not have selection lag so. You will see if this works.

PackageExplorer/PackageViewer.xaml
mangelovic: Did not applied your change. I did not manage to find situation, when CONTROL click did not work for deselection of folder. Anyway, adding `IsHitTestVisible="False"` made another problem. It may change another functionalities too, but I tested it on building of new package. 

When I rightclicked on folder and wanted to "Add New File", it was not added into clicked folder, but outside of it.

campersau: Did you try deselecting by clicking on the folder name (text) instead of the icon? Then it does not work for me. I tested it with the build from your latest commit https://dev.azure.com/onovotny/GitBuilds/_build/results?buildId=2871&view=artifacts

PackageExplorer/Controls/MultiSelectTreeView.cs
campersau: Looks like 5ms is small enough for holding the arrow keys down without lagging.

PackageExplorer/PackageViewer.xaml
mangelovic: I checked it out and it occurs on trying to deselect last selected folder. I will try to solve it out. But it is a bit strange that clicking on icon works, but not at text.

304NotModified:
closing a inactive for a while

Issue #838, opened 2019-10-01
Support for displaying deprecated packages


onovotny:
NuGet's metadata can display package deprecation information. We should display it in a similar way in the Package Explorer UX

https://devblogs.microsoft.com/nuget/deprecating-packages-on-nuget-org/

![image](https://user-images.githubusercontent.com/1427284/65925127-8a4ae900-e3bd-11e9-9238-7bcb2d5df299.png)

loic-sharma:
You should be able to use the package metadata resource (aka "registration) to grab a package's deprecation information: https://docs.microsoft.com/en-us/nuget/api/registration-base-url-resource#package-deprecation

campersau:
For me visual studio does not show deprecated versions in the list view. Only in the package version details:
![image](https://user-images.githubusercontent.com/4009570/68527872-487e5e00-02ec-11ea-90a4-08519028123e.png)

Is that correct? Was it removed or will it be available in the future?

PR #855, opened 2019-11-09
show deprecated packages in package list details

campersau:
Currently only showing in the package version details, because this is how VS does it on my machine.
See https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/issues/838#issuecomment-552091187

I can also add it to the package list, which would look like this:
image

fixes https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/issues/838

### DIFF
--- a/Core/Packages/DeprecationInfo.cs
+++ b/Core/Packages/DeprecationInfo.cs
@@ -1,0 +1,24 @@
+﻿using System.Collections.Generic;
+using NuGet.Versioning;
+
+namespace NuGetPe
+{
+    public class DeprecationInfo
+    {
+        public string? Message { get; set; }
+#pragma warning disable CS8618 // Non-nullable field is uninitialized. Consider declaring as nullable.
+        public IEnumerable<string> Reasons { get; set; }
+#pragma warning restore CS8618 // Non-nullable field is uninitialized. Consider declaring as nullable.
+        public AlternatePackageInfo? AlternatePackageInfo { get; set; }
+    }
+
+    public class AlternatePackageInfo
+    {
+#pragma warning disable CS8618 // Non-nullable field is uninitialized. Consider declaring as nullable.
+        public string Id { get; set; }
+#pragma warning restore CS8618 // Non-nullable field is uninitialized. Consider declaring as nullable.
+#pragma warning disable CS8618 // Non-nullable field is uninitialized. Consider declaring as nullable.
+        public VersionRange Range { get; set; }
+#pragma warning restore CS8618 // Non-nullable field is uninitialized. Consider declaring as nullable.
+    }
+}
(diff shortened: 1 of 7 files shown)


onovotny:
Looks great in the list view as well!

Does the depreciation data show a message and replacement version? If so, could be good to show the message as a tooltip under the shield in the list view and next to it in the details with a link to the replacement package.

Core/Packages/DeprecationInfo.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System.Collections.Generic;
+using NuGet.Versioning;
+
+namespace NuGetPe
+{
+    public class DeprecationInfo
+    {
+        public string? Message { get; set; }
+        public IEnumerable<string>? Reasons { get; set; }
```
onovotny: I don't think IEnumerable's should ever be nullable. At best, it can return an empty enumerable.

Core/Packages/DeprecationInfo.cs
```diff
@@ -0,0 +1,18 @@
...
+        public AlternatePackageInfo? AlternatePackageInfo { get; set; }
+    }
+
+    public class AlternatePackageInfo
+    {
+        public string? Id { get; set; }
```
onovotny: Can Id be null if there's an alternate package?

Core/Packages/DeprecationInfo.cs
```diff
@@ -0,0 +1,18 @@
...
+        public VersionRange? Range { get; set; }
```
onovotny: How about the range? Is that optional?

PackageViewModel/PackageChooser/PackageInfoViewModel.cs
```diff
@@ -283,6 +318,11 @@ private void OnCancelDownload()
             }
         }
 
+        private void OnOpenAlternatePackage(string packageId)
+        {
+            _parentViewModel.SearchCommand.Execute("id:" + packageId);
```
onovotny: We should probably log a telemetry event for Opening an alternate package like we do the other user actions/events.

Core/Packages/DeprecationInfo.cs
campersau: Yes, I did it mostly because of the nullable warnings.

I can add
```c#
#pragma warning disable CS8618 // Non-nullable field is uninitialized. Consider declaring as nullable.
````
around them.

PackageViewModel/PackageChooser/PackageInfoViewModel.cs
campersau: Tracked as: `PackageInfoViewModel_OpenAlternatePackage`

Core/Packages/DeprecationInfo.cs
onovotny: Yeah, I wish the code analysis would be better for DTO's. The pragma works until they improve it.

campersau:
Screenshot with deprecation message:
image

Screenshot with alternate package:
image

onovotny:
LGTM, merge away when ready

Issue #820, opened 2019-08-26
Support Icons within the package


onovotny:
NuGet is moving to include icons within the package. We need to support that:

Spec: https://github.com/NuGet/Home/wiki/Packaging-Icon-within-the-nupkg

We should also have a validation rule for ensuring the icon is actually within the package (error on save?) as that would fail push anyway.

onovotny:
FakeItEasy 5.4.0 supports this. Looks like the feed metadata turns the url into a chached public link, so nothing to do for that view:
![image](https://user-images.githubusercontent.com/1427284/68531712-4a8fef00-02e3-11ea-994e-2ad3908a56d9.png)

We need to determine the best way to display it in the package view, indexing into the package stream so we don't need to extract it.

campersau:
I'm currently working on it. Indexing works fine also for manual package creation:  https://github.com/campersau/NuGetPackageExplorer/blob/081eaadc5f8fe55b5264de014ac2f17463e663b1/PackageExplorer/Converters/PackageIconConverter.cs

PR #856, opened 2019-11-09
Package icon

campersau:
Package Viewer now displays either the `Icon` or the `IconUrl`. Before and after:
image

In package editing the user can select an existing icon from the package content or specify an URL by hand:
image

fixes https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/issues/820

### DIFF
--- a/PackageExplorer/Converters/PackageIconConverter.cs
+++ b/PackageExplorer/Converters/PackageIconConverter.cs
@@ -20,7 +20,7 @@ namespace PackageExplorer
                 {
                     foreach (var file in package.RootFolder.GetFiles())
                     {
-                        if (file.Path == metadata.Icon)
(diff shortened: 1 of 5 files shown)


PackageExplorer/Converters/PackageIconConverter.cs
```diff
@@ -0,0 +1,55 @@
...
+        {
+            if (values?.Length > 1 && values[0] is PackageViewModel package && values[1] is string str && !string.IsNullOrEmpty(str))
+            {
+                var metadata = package.PackageMetadata;
+
+                if (!string.IsNullOrEmpty(metadata.Icon))
+                {
+                    foreach (var file in package.RootFolder.GetFiles())
+                    {
+                        if (file.Path == metadata.Icon)
```
onovotny: This should probably be a case-insensitive ordinal compare.

PR #900, opened 2020-03-19
Net5p1

clairernovotny:


campersau:
There is another `netcoreapp3.1` reference here:
https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/blob/60480a8c34fa6a5072cf96d20599502992413734/azure-pipelines.yml#L128

PR #926, opened 2020-04-04
Support sibling snupkg file

### LINKED ISSUE #925: Opening local .nupkg doesn't pick up sibling .snupkg
**Type**:

- Feature request

**NPE version**: NuGet Package Explorer - Store - X64 - (5.6.40+0baf79cb07)

**OS version**: Windows 10 - 1909

Thanks for this application! It highlighted my package [`MetadataExtractor/2.3.0`](https://www.nuget.org/packages/MetadataExtractor/2.3.0) as not being configured correctly for SourceLink:

image

So I'm now working locally to fix it.  Building produces both nupkg/snupkg files, and when I load the nupkg in NuGet Package Explorer I see:

image

I'm confused why my published package (2.3.0) says there's no SourceLink for three of the four targets' DLLs, as I am referencing SourceLink and believe it's configured appropriately. I would like to make changes locally and use NPE to verify correctness before publishing the package, but I seem to be stuck on NPE finding the symbols in the supkg file.

drewnoakes:
Fixes #925.

When opening a `nupkg` file from disk, if its symbols are in a `snupkg` file next to it, it would be flagged as having no symbols. This commit tests for such a file, using it if possible, before returning to the prior behaviour of checking nuget.org for packages identified as coming from there.

### DIFF
--- a/PackageViewModel/SymbolValidation/SymbolValidator.cs
+++ b/PackageViewModel/SymbolValidation/SymbolValidator.cs
@@ -258,51 +258,69 @@ namespace PackageExplorerViewModel
 
             var requireExternal = false;
             // See if any pdb's are missing and check for a snupkg on NuGet.org. 
-            if (noSymbols.Count > 0 && _publishedOnNuGetOrg)
+            if (noSymbols.Count > 0)
             {
-                // try to get on NuGet.org
-                // https://www.nuget.org/api/v2/symbolpackage/Newtonsoft.Json/12.0.3 -- Will redirect               
-
-                
                 try
                 {
+                    // try to find a sibling snupkg file locally
+                    var snupkgFilePath = Path.ChangeExtension(_packageViewModel.PackagePath, ".snupkg");
+                    var symbolsFilePath = Path.ChangeExtension(_packageViewModel.PackagePath, ".symbols.nupkg");
+                    if (File.Exists(snupkgFilePath))
+                    {
+                        ReadSnupkgFile(snupkgFilePath);
+                    }
+                    else if (File.Exists(symbolsFilePath))
+                    {
+                        ReadSnupkgFile(symbolsFilePath);
+                    }
+                    else if (_publishedOnNuGetOrg)
+                    {
+                        // try to get on NuGet.org
+                        // https://www.nuget.org/api/v2/symbolpackage/Newtonsoft.Json/12.0.3 -- Will redirect
+
 #pragma warning disable CA2234 // Pass system uri objects instead of strings
-                    var response = await _httpClient.GetAsync($"https://www.nuget.org/api/v2/symbolpackage/{_package.Id}/{_package.Version.ToNormalizedString()}").ConfigureAwait(false);
+                        var response = await _httpClient.GetAsync($"https://www.nuget.org/api/v2/symbolpackage/{_package.Id}/{_package.Version.ToNormalizedString()}").ConfigureAwait(false);
 #pragma warning restore CA2234 // Pass system uri objects instead of strings
 
-                    if (response.IsSuccessStatusCode) // we'll get a 404 if none
-                    {
-                        requireExternal = true;
-
-                        using var getStream = await response.Content.ReadAsStreamAsync();
-                        using var tempFile = new TemporaryFile(getStream, ".snupkg");
-                        using var package = new ZipPackage(tempFile.FileName);
-
-                        // Look for pdb's for the missing files
-                        var dict = package.GetFiles().ToDictionary(k => k.Path);
-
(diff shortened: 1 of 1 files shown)


PackageViewModel/SymbolValidation/SymbolValidator.cs
```diff
@@ -258,51 +258,68 @@ private async Task CalculateValidity(IReadOnlyList<PackageFile> files)
...
-            if (noSymbols.Count > 0 && _publishedOnNuGetOrg)
+            if (noSymbols.Count > 0)
             {
-                // try to get on NuGet.org
-                // https://www.nuget.org/api/v2/symbolpackage/Newtonsoft.Json/12.0.3 -- Will redirect               
-
-                
                 try
                 {
+                    if (File.Exists(_packageViewModel.PackagePath))
```
clairernovotny, line 265: This if/else should probably check for the snupkg locally and fallback to nuget.org. The case may be if the nuget file is downloaded locally from nuget and then opened.

drewnoakes: That was my intention. My mistake was assuming the `PackagePath` would be invalid (and this would return `false`) for packages from feeds. In actuality it's in a temp folder, e.g. `%LocalAppData%\Temp\tmp94F5.tmp`. The odds of a temp file being called e.g. `tmp94F5.snupkg` seem slim to me. I'll push an update now.

PackageViewModel/SymbolValidation/SymbolValidator.cs
```diff
@@ -258,51 +258,66 @@ private async Task CalculateValidity(IReadOnlyList<PackageFile> files)
...
-                // try to get on NuGet.org
-                // https://www.nuget.org/api/v2/symbolpackage/Newtonsoft.Json/12.0.3 -- Will redirect               
-
-                
                 try
                 {
+                    // try to find a sibling snupkg file locally
+                    var directoryName = Path.GetDirectoryName(_packageViewModel.PackagePath) ?? string.Empty;
+                    var fileNameWithoutExtension = Path.GetFileNameWithoutExtension(_packageViewModel.PackagePath);
+                    var snupkgFilePath = Path.Combine(directoryName, fileNameWithoutExtension + ".snupkg");
```
clairernovotny, line 268: Nit: there’s a method Path.ChangeFileExtension that does these few lines

drewnoakes: Much nicer, thanks!

clairernovotny:
I’ll test and merge tomorrow, thanks!

PackageViewModel/SymbolValidation/SymbolValidator.cs
```diff
@@ -258,51 +258,64 @@ private async Task CalculateValidity(IReadOnlyList<PackageFile> files)
...
-                    var response = await _httpClient.GetAsync($"https://www.nuget.org/api/v2/symbolpackage/{_package.Id}/{_package.Version.ToNormalizedString()}").ConfigureAwait(false);
+                        var response = await _httpClient.GetAsync($"https://www.nuget.org/api/v2/symbolpackage/{_package.Id}/{_package.Version.ToNormalizedString()}").ConfigureAwait(false);
 #pragma warning restore CA2234 // Pass system uri objects instead of strings
 
-                    if (response.IsSuccessStatusCode) // we'll get a 404 if none
-                    {
-                        requireExternal = true;
+                        if (response.IsSuccessStatusCode) // we'll get a 404 if none
+                        {
+                            requireExternal = true;
```
clairernovotny, line 282: We need to set `requireExternal` to true for the local `snupkg` case as well. This is because it'll ultimately require a symbol server to work.

clairernovotny:
Oh, one other thought -- as long as we're checking for `.snupkg` sibling files, we should also check for `.symbols.nupkg`.

PackageViewModel/SymbolValidation/SymbolValidator.cs
```diff
@@ -258,51 +258,69 @@ private async Task CalculateValidity(IReadOnlyList<PackageFile> files)
...
                 try
                 {
+                    // try to find a sibling snupkg file locally
+                    var snupkgFilePath = Path.ChangeExtension(_packageViewModel.PackagePath, ".snupkg");
+                    var symbolsFilePath = Path.ChangeExtension(_packageViewModel.PackagePath, ".symbols.nupkg");
+                    if (File.Exists(snupkgFilePath))
+                    {
+                        ReadSnupkgFile(snupkgFilePath);
+                    }
+                    if (File.Exists(symbolsFilePath))
```
clairernovotny, line 272: should be `else if`

Issue #1000, opened 2020-05-28
Why does PDB checksum validation fail the Source Link and Deterministic checks?


bording:
**Type** Question / feature request
**NPE version**: 5.7.27
**OS version**: Windows 10 1909 - 18363.836
**Installed from**: Microsoft Store

Ever since NuGet Package Explorer added PDB checksum validation, our packages have been showing up as not having Source Link support despite being properly indexed:

![image](https://user-images.githubusercontent.com/753669/83191815-a5aff480-a102-11ea-8b8f-b61394bfcfd5.png)

It's failing because of the checksum validation: 

![image](https://user-images.githubusercontent.com/753669/83191938-ca0bd100-a102-11ea-981c-33e0498f02e9.png)

The package used in the screen shots: [NServiceBus 7.3.0](https://www.nuget.org/packages/NServiceBus/7.3.0)

However, despite that, Source Link works fine on our packages, letting you step in to code as expected.

We do use Fody for post-processing of the assemblies, and it appears that it is removing the `PdbChecksum` `DebugDirectoryEntry` from the PDB when it does. While I am going to see if there's a way to get Fody to write a valid checksum value into the PDB, I think it makes sense to change the way NuGet Package Explorer handles this scenario.

What does the checksum validation have to do with Source Link? It seems to me that it should be a separate entry if you want to report that.

I also think that the deterministic check should not by skipped in this scenario. It currently fails because of "Missing Symbols", which also isn't true in this case.

clairernovotny:
The checksum is there to ensure that the pdb matches the dll, which is important for debugging and traceability. The fact that it works is likely an accident....  @tmat, any thoughts here?

bording:
In this case, it's not that the checksum doesn't match, it's the PDB doesn't have a checksum entry at all.

clairernovotny:
I believe that's an invalid PDB then.

bording:
> I believe that's an invalid PDB then.

Except it works fine in every case that I can think of. If the VS debugger can use it successfully, Source Link works, etc. what is invalid about it?

clairernovotny:
I just talked to @tmat. In this case, when the pdb is in the nupkg, it looks like the checksum isn't needed. I can fix the logic to account for that.

*However* snupkg requires the checksum for things to match and work so there's still an issue that needs to be fixed in Fody.

clairernovotny:
@simoncropp you may want to take a look at this 👆

[...]

PR #1001, opened 2020-05-28
Don't check the checksum for pdbs alongside the dll

clairernovotny:
Fixes for #1000 

Note that there are still untracked files here, and `ContinuousIntegrationBuild` needs to be set to true.

image


clairernovotny:
@bording 👆

bording:
Thanks for the quick change!

I'm aware of the other issues. I'm in the process of fixing all that up for our build system when I realized there was something odd going on with NuGet Package Explorer.

Issue #919, opened 2020-03-30
Package analyzer CLI (or consumable analysis library)


tillig:
**Type**: Feature request
**OS version**: Mac OS 10.15.4 (yeah, I'll explain)

I've been doing a lot of my development/packaging on MacOS and Linux platforms since .NET Core came out. Unfortunately, that means no NuGet Package Explorer for me, which isn't bad; I can use Parallels or something like that if it becomes necessary. Note my majority use case is browsing existing packages, not creating or publishing packages.

I noticed some of the [pretty sweet 'NuGet package analysis' features tweeted](https://twitter.com/clairernovotny/status/1244007715778834432) by @clairernovotny and it got me a thinking... it would be really cool to be able to run at least the package analysis part from a command line. That way I could check a built package without having to fire up a VM... but also, I could add something like that as part of my CI build and check the results there, failing if things don't check out.

I understand this is a bit outside the scope of NuGet Package Explorer, but it occurs to me that if the bits that do the analysis were in some consumable package outside the UI it would be easy enough to wrap them in a global `dotnet` tool and call it a day.

The "simplest thing that could possibly work" would be something like writing out a JSON-formatted report of the health parts of the analysis:

```json
{
  "health": {
    "signature": {
      "valid": true
    },
    "sourceLink": {
      "valid": true
    },
    "deterministic": {
      "valid": false,
      "messages": [
        "Package is non-deterministic.",
        "Ensure the following properties are enabled on CI builds..."
      ]
    }
  }
}
```

Admittedly, that may not be the best report schema. Maybe it needs to be something more like what comes out of a Roslyn analyzer with a warning code and a message, such that only issues are reported and successes are omitted.

```json
[
  { "code": "NGPE10003", "message": "Package is non-deterministic." }
]
```

Regardless, the idea is that it'd be really cool to be able to access that analysis bit outside the UI in a .NET Core app so it could be used cross-platform.

clairernovotny:
I agree this would be awesome. This would require some refactoring though as we're currently depending on `PackageFile` in the ViewModel layer as it handles the path/tree parsing and associations. 

I'd be happy for someone to jump in on this. `Core` can probably be used as-is as it's not ViewModel.

[...]

PR #1010, opened 2020-06-03
Add a CLI symbol validator

0xced:
This is a work in progress of a CLI symbol validator as discussed in #919.

I'm opening this pull request so that people interested can follow the progress but note that for now, this is just some exploratory refactoring.

### DIFF
--- a/Core/Core.csproj
+++ b/Core/Core.csproj
@@ -14,6 +14,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Types\Types.csproj" />
+    <PackageReference Include="AuthenticodeExaminer" Version="0.3.0" />
     <PackageReference Include="Microsoft.ApplicationInsights" Version="2.14.0" />
     <PackageReference Include="Microsoft.DiaSymReader.Converter" Version="1.1.0-beta2-20216-01" />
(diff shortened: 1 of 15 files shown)


Core/Core.csproj
```diff
@@ -14,13 +14,11 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Types\Types.csproj" />
-    <PackageReference Include="AppInsights.WindowsDesktop" Version="2.13.1" />
+    <PackageReference Include="Microsoft.ApplicationInsights" Version="2.14.0" />
```
clairernovotny, line 18: We can't use this package as it doesn't contain the correct telemetry. AppInsights.WindowsDesktop should support other platforms, the key part is that it buffers and persists telemetry.

clairernovotny:
There isn't a huge dependency on the ViewModel types -- except for the directory path parsing to get sibling folders. That code may be refactored to pull some bits down into core.

clairernovotny:
@0xced just wondering how this is coming along?

clairernovotny:
Closing as the work was finished in #1210. Thank you so much for getting this started!

PR #1028, opened 2020-06-12
Don't focus on ItemsControl in PdbInfoViewer / PdbSourcesViewer

### LINKED ISSUE #1027: PdbInfo scroll box is centered instead of left
Looks like the control starts centered instead of left. There's also a brief pause/delay loading the tab and I'm not sure why. The data should already be parsed and available.

Can repro with package attacked to #1024 

image

@campersau any ideas?

campersau:
fixes #1027

### DIFF
--- a/PackageExplorer/Controls/PdbInfoViewer.xaml
+++ b/PackageExplorer/Controls/PdbInfoViewer.xaml
@@ -35,6 +35,7 @@
             <RowDefinition Height="Auto" />
         </Grid.RowDefinitions>
 
+        <!-- IsTabStop="True" here is to prevent scrolling down to the first focusable element which would be the first cell in the DataGrid -->
         <TextBox Text="Type:" Style="{StaticResource Key}" IsTabStop="True" />
         <TextBox
             Grid.Column="1"            
--- a/PackageExplorer/Controls/PdbInfoViewer.xaml
+++ b/PackageExplorer/Controls/PdbInfoViewer.xaml
@@ -35,7 +35,7 @@
             <RowDefinition Height="Auto" />
         </Grid.RowDefinitions>
 
-        <TextBox Text="Type:" Style="{StaticResource Key}" />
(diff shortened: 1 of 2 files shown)


PackageExplorer/Controls/PdbInfoViewer.xaml
```diff
@@ -63,7 +63,7 @@
             Grid.Column="1"
             Visibility="{Binding CompilerFlags, Converter={StaticResource NullToVisibilityConverter}}"
             ItemsSource="{Binding CompilerFlags}"
-            MaxHeight="250">
```
campersau, line 66: @clairernovotny was there a reason why you did set a `MaxHeight`?

PackageExplorer/Controls/PdbInfoViewer.xaml
```diff
@@ -35,7 +35,7 @@
             <RowDefinition Height="Auto" />
         </Grid.RowDefinitions>
 
-        <TextBox Text="Type:" Style="{StaticResource Key}" />
+        <TextBox Text="Type:" Style="{StaticResource Key}" IsTabStop="True" />
```
campersau, line 38: This is a workaround to not auto focus the first cell in the `DataGrid` which would scroll it into view...

PackageExplorer/Controls/PdbInfoViewer.xaml
clairernovotny: The idea was that if the compiler list became too long, it'd scroll. That way the references wouldn't be too far down. Not sure on the best UX for the tab as a whole.

PackageExplorer/Controls/PdbInfoViewer.xaml
clairernovotny: Lets just add a comment in the code so we remember why we did that down the road :)

campersau: Done.

PR #1059, opened 2020-07-01
Support parsing standalone pdbs which is useful for snupkg

campersau:
By doing some `null` checks.

Example when viewing details for pdb inside of snupkg:

Before:
image

After:
image

### DIFF
--- a/Core/AssemblyMetadata/AssemblyDebugParser.cs
+++ b/Core/AssemblyMetadata/AssemblyDebugParser.cs
@@ -48,14 +48,16 @@ namespace NuGetPe.AssemblyMetadata
                 pdbStream.Position = 0;
             }
 
-            
 
-            _readerProvider = MetadataReaderProvider.FromPortablePdbStream(inputStream);            
+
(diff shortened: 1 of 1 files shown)


clairernovotny:
Snupkgs should still have the dll's it though? The main issue is that we try to convert Windows PDB's to Portable so we can do the rest uniformly. Windows PDB's need to have the PE file passed in too.

This also works for .symbols.nupkg files, so while technically snupkg does not support windows pdb's, .symbols.nupkg does.

campersau:
When I use
```xml
<IncludeSymbols>true</IncludeSymbols>
<SymbolPackageFormat>snupkg</SymbolPackageFormat>
```
only the PDB is inside the `.snupkg`
image

Yes, and I haven't changed the path for Windows PDBs. So now it is possible to get more info about portable PDBs without the corresponding DLL.
https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/blob/f486e94454d0e79ff408d65afcf4325ab32b5709/Core/AssemblyMetadata/AssemblyDebugParser.cs#L25-L28

Before a NRE would be thrown for portable PDBs (and catched).

Issue #1097, opened 2020-08-31
Opening symbol package crashes the explorer. System.Windows.Markup.XamlParseException


hoyosjs:
**Type**: Bug
**NPE version**: 5.7.145+8047943308
**OS version**: Windows 10 - 2004/19041.450
**Installed from**: Microsoft Store

In case of a BUG:
- Opening some packages crashes the explorer. I managed to take a full dump, looking at it there's the following exception.
```
(130e8.13390): CLR exception - code e0434352 (first/second chance not available)
CLR exception type: System.Windows.Markup.XamlParseException
    "Number of entries expected in End Of Central Directory does not correspond to number of entries in Central Directory."
```
<details>
<summary>Details from !analyze</summary>

```
*******************************************************************************
*                                                                             *
*                        Exception Analysis                                   *
*                                                                             *
*******************************************************************************

DEBUG_FLR_EXCEPTION_CODE(80131501) and the ".exr -1" ExceptionCode(e0434352) don't match

KEY_VALUES_STRING: 1

    Key  : Analysis.CPU.mSec
    Value: 2046

    Key  : Analysis.Elapsed.mSec
    Value: 2438

    Key  : Analysis.Init.CPU.mSec
    Value: 13327

    Key  : Analysis.Init.Elapsed.mSec
    Value: 1647012

    Key  : Analysis.Memory.CommitPeak.Mb
    Value: 970

    Key  : CLR.Engine
    Value: CORECLR

    Key  : CLR.Exception.System.IO.InvalidDataException._message
    Value: Number of entries expected in End Of Central Directory does not correspond to number of entries in Central Directory.

    Key  : CLR.Exception.System.Windows.Markup.XamlParseException._lineNumber
    Value: 0n00000000000

    Key  : CLR.Exception.System.Windows.Markup.XamlParseException._linePosition
    Value: 0n00000000000

    Key  : CLR.Exception.System.Windows.Markup.XamlParseException._message
    Value: Number of entries expected in End Of Central Directory does not correspond to number of entries in Central Directory.

    Key  : CLR.Version
    Value: 5.0.20.40711

    Key  : Memory.Job.PrivateCommit.MB
    Value: 125

    Key  : Memory.Job.PrivateLimit.MB
    Value: 0

    Key  : Memory.Job.SharedCommit.MB
    Value: 22

    Key  : Memory.Job.TotalLimit.MB
    Value: 0

[...]

PR #1106, opened 2020-09-13
catch exceptions when loading images from zip

campersau:
I couldn't reproduce it during debugging, only when running without debugging and very rarely.
So I just added a try catch to prevent the crash...

fixes https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/issues/1097

### DIFF
--- a/PackageExplorer/Converters/PackageIconConverter.cs
+++ b/PackageExplorer/Converters/PackageIconConverter.cs
@@ -28,12 +28,18 @@ namespace PackageExplorer
                     {
                         if (string.Equals(file.Path, iconPath, StringComparison.OrdinalIgnoreCase))
                         {
-                            var image = new BitmapImage();
-                            image.BeginInit();
-                            image.CacheOption = BitmapCacheOption.OnLoad;
-                            image.StreamSource = file.GetStream();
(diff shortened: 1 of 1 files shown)


PackageExplorer/Converters/PackageIconConverter.cs
```diff
@@ -28,12 +28,18 @@ public class PackageIconConverter : IMultiValueConverter
...
-                            image.EndInit();                            
-                            return image;
+                            // catch potential exceptions during image loading from zip file
+                            // see 1097
+                            try
+                            {
+                                var image = new BitmapImage();
+                                image.BeginInit();
+                                image.CacheOption = BitmapCacheOption.OnLoad;
+                                image.StreamSource = file.GetStream();
```
clairernovotny, line 38: Is this really going to catch it though? It depends on where the exception happens since this by this point it's just opening the stream. I think load happens by the BitmapImage happens later.

Is there a call stack that shows the crash directly?

campersau: See https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/issues/1097#issuecomment-691667911

PR #1114, opened 2020-09-19
Record package id's for public packages

clairernovotny:


Core/Utility/DiagnosticsClient.cs
```diff
@@ -56,6 +56,29 @@ public static void TrackException(Exception exception, IDictionary<string, strin
...
+                    };
+
+                TrackException(exception, info);
+            }
+            else
+            {
+                TrackException(exception);
+            }
+
+            _client.TrackException(exception, properties, metrics);
```
campersau, line 79: Do the above `TrackException`s also call `_client.TrackException` so you they would be reported twice?

Issue #1207, opened 2021-01-20
[Feature] Commandline report on package health and analyze results


quasarea:
- Feature request
Be able to get results of health check and package analyse via commandline 

- Why do we need it?
To provide means for showing package quality in devops pipelines

- Please provide an use case.
So I can automatically add pull request comment that package assemblies are non deterministic and located outside known folders

clairernovotny:
@0xced seems to be working on it https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/pull/1010 but I'm not sure what the status is?

I would like this as well.

0xced:
I still have a lot of duplicated code that I want to deduplicate before turning the draft pull request into a final pull request. Not sure when I'll get there though…

clairernovotny:
Is this something someone else could finish up? What code is duplicated and what still needs to be done?

quasarea:
could output report in a sarif (json) format, so it stays inline with other tools like devskim, binskim, etc.

PR #1210, opened 2021-01-24
Cross platform cli validator

### LINKED ISSUE #919: Package analyzer CLI (or consumable analysis library)

clairernovotny:
Fixes #919, #1207.

Based on @0xced's initial work


0xced:
Thank you very much for finishing the remaining work, Claire!

PR #1213, opened 2021-01-25
dotnet-validate: Add initial support for glob patterns in local package validation...

danielcweber:
... with the help of Microsoft.Extensions.FileSystemGlobbing. Can handle fully qualified and relative paths, but there might be edge cases.

PR #1229, opened 2021-02-02
Fix an error introduced in #1213...

danielcweber:
…where the relative file name would not be combined with the (root-) directory determined earlier.


0xced:
Looks good to me, that's exactly how I fixed this issue on both macOS and Windows.

campersau (approved):

Issue #1095, opened 2020-08-26
Support Embedded Readme


campersau:
**Type**: Feature request

- [ ] Support `PackageReadmeFile` in `PackageMetadataEditor` https://github.com/NuGet/NuGet.Client/pull/3575
- [ ] Support `ReadmeUrl` in `PackageChooserDialog` https://github.com/NuGet/NuGet.Client/pull/3583

Waiting for https://github.com/NuGet/NuGet.Client/pull/3634

PR #1266, opened 2021-03-16
Add support for NuGet Readme

campersau:
Fixes https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/issues/1095

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -4,7 +4,11 @@
     <UseWpf>true</UseWpf>
     <UseWindowsForms>true</UseWindowsForms>
   </PropertyGroup>
-  
+
+  <PropertyGroup Condition="'$(TargetFramework)' == 'netcoreapp3.1'">
+    <NuGetDependencyVersion>5.10.0-xprivate.60024</NuGetDependencyVersion>
+  </PropertyGroup>
+
   <PropertyGroup>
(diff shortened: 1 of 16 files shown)


Directory.Build.targets
```diff
@@ -4,10 +4,6 @@
     <UseWpf>true</UseWpf>
     <UseWindowsForms>true</UseWindowsForms>
   </PropertyGroup>
-
-  <PropertyGroup Condition="'$(TargetFramework)' == 'netcoreapp3.1'">
```
campersau, line 8: @clairernovotny was this a special build related to some signature stuff?
I have removed the `netcoreapp3.1` target for now.

clairernovotny: We need the .NET Core App 3.1 version as other things are consuming the Core nuget package for symbol validation. The Core version is using a private build of NuGet that does the signature validation.

clairernovotny (changes requested):
Looks great other than the .NET Core 3.1 removal. If you could please add that back, I can merge.

clairernovotny:
Could you also please bump the `version.json` to be 5.10? I've been keeping the minor in sync with NuGet to get a sense of what features are where.

campersau:
@clairernovotny I have bumped the version to `5.10`.

PR #1281, opened 2021-05-26
Remove duplicate MakeSeekable method from SymbolValidator

campersau:
Implementation looks identical to:
https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/blob/6a2d01cb3f95734ef2203234851e782043172581/Core/Utility/StreamUtility.cs#L16-L35

Also `StreamUtility.MakeSeekable` was already used here:
https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/blob/6a2d01cb3f95734ef2203234851e782043172581/Core/SymbolValidation/SymbolValidator.cs#L527

### DIFF
--- a/Core/SymbolValidation/SymbolValidator.cs
+++ b/Core/SymbolValidation/SymbolValidator.cs
@@ -539,7 +539,7 @@ namespace NuGetPe
             List<IFile> nonReproducible,
             bool validateChecksum)
         {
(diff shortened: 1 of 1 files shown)


clairernovotny:
strange, not sure why that was even there 🤷‍♀️

PR #1294, opened 2021-06-22
feat: Add Uno Platform targets

jeromelaban:
This PR adds the support for NPE on the Web, through Uno Platform and WebAssembly.

A few notes on this PR:
- Other heads are also present, like macOS and Skia heads, but are not yet ready for distribution.
- The UWP head folder is present for the support on Skia and WebAssembly heads, and will be migrated to WinUI 3 when ready.
- The build pipeline is being worked on, notably to get the Azure Static WebApp running properly, particularly with the automated PR review sites generation.


azure-pipelines.yml
```diff
@@ -174,7 +177,7 @@ stages:
       displayName: Build Uno Apps
       inputs:
         solution: $(build.sourcesdirectory)/NuGetPackageExplorer-uno.slnf
-        msbuildArguments: /r /p:Configuration=Release /p:UNO_BUILD_ONLY=true /bl:$(Build.ArtifactStagingDirectory)\Logs\$(ReleaseChannel).binlog
+        msbuildArguments: /r /p:Configuration=Release /p:UNO_BUILD_ONLY=true /p:NpeAiInstrumentationKey=$(NPE_AI_INSTRUMENTATIONKEY) /bl:$(Build.ArtifactStagingDirectory)\Logs\$(ReleaseChannel).binlog
```
clairernovotny, line 180: There's already a variable for this in the pipeline, though it's not in its own variable group yet:

https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/blob/main/azure-pipelines.yml#L116-L121

clairernovotny: if we can do it in a script instead of a build variable, it keeps it out of the binlog too.

jeromelaban: I don't think we'll be able to put it in a script, since it has to be an msbuild variable to become an wasm environment variable.

clairernovotny: What if we put a placeholder `REPLACE_ME_WITH_THE_KEY` in the msbuild variable and replace it in the generated output?

jeromelaban: The output is compressed by the tooling, we'd need to re-generate the compressed file as well, or delete it so the uncompressed file takes over. Still, the key is easily readable from the live app, so changing this won't make it more secret.

clairernovotny: Gotcha, so be it.

.github/workflows/azure-static-web-app-pr-validation.yml
```diff
@@ -0,0 +1,67 @@
...
+          
+      - name: Setup dotnet 3.1.409
+        uses: actions/setup-dotnet@v1.7.2
+        with:
+          dotnet-version: '3.1.409'
+          
+      - name: Setup dotnet 5.0.103
+        uses: actions/setup-dotnet@v1.7.2
+        with:
+          dotnet-version: '5.0.103'
```
clairernovotny, line 32: The dotnet versions should be 3.x/5.x/6.x to ensure they get the latest security fixes

.github/workflows/azure-static-web-app-pr-validation.yml
```diff
@@ -0,0 +1,67 @@
...
+          
+      - run: |
+          cd NuGetPackageExplorer/NuGetPackageExplorer.Wasm
+          dotnet build -c Release /p:UNO_BUILD_ONLY=true /p:NpeAiInstrumentationKey=${{secrets.NPE_AI_INSTRUMENTATIONKEY}}
```
clairernovotny, line 36: We don't need the AI key for PRs

filipnavara:
Thanks! <3

PR #1320, opened 2021-07-12
fix: Adjust instrumentation key variable

jeromelaban:
Adjust the NPE_AI_INSTRUMENTATIONKEY variable to match what's needed by the javascript version of AppInsights.

@clairernovotny We can do this, or make a different variable that uses the required formatting. Which one do you prefer ?


jeromelaban:
I'll merge it as-is to avoid losing data, We'll adjust later if needed.

PR #1345, opened 2021-08-13
[wasm] download selected file on double click

campersau:
Will automatically get more awesome with https://github.com/unoplatform/uno/pull/6619

### DIFF
--- a/Uno/NuGetPackageExplorer.Shared/InspectPackage.xaml
+++ b/Uno/NuGetPackageExplorer.Shared/InspectPackage.xaml
@@ -771,6 +771,7 @@
 						   behavior:TreeViewBehavior.EnableSelectedItemBinding="True"
 						   behavior:TreeViewBehavior.SelectedItem="{Binding SelectedContent, Mode=TwoWay}"
 						   behavior:TreeViewBehavior.AutoToggleItemExpansion="True"
+						   behavior:TreeViewBehavior.DoubleClickCommand="{Binding DoubleClickCommand}"
 						   ItemTemplateSelector="{StaticResource FSObjectTemplateSelector}"
 						   ItemContainerStyle="{StaticResource FileExplorerTreeViewItemStyle}"
 						   SelectionMode="Single" />
--- a/Uno/NugetPackageExplorer.Legacy/Presentation/Content/InspectPackageViewModel.cs
+++ b/Uno/NugetPackageExplorer.Legacy/Presentation/Content/InspectPackageViewModel.cs
@@ -34,6 +34,7 @@ using Uno.Logging;
 
 using Windows.ApplicationModel.Core;
 using Windows.Storage;
+using Windows.Storage.Pickers;
 using Windows.UI.Xaml;
 using Windows.UI.Xaml.Controls;
 
@@ -68,6 +69,8 @@ namespace NupkgExplorer.Presentation.Content
             get => GetProperty<string?>();
             set => SetProperty(value);
         }
+
+        public ICommand DoubleClickCommand => GetCommand(DoubleClick);
 
         public ICommand CloseDocumentCommand => GetCommand(CloseDocument);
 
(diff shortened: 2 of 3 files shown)


Uno/NuGetPackageExplorer.Shared/InspectPackage.xaml
```diff
@@ -787,6 +787,7 @@
 						   behavior:TreeViewBehavior.EnableSelectedItemBinding="True"
 						   behavior:TreeViewBehavior.SelectedItem="{Binding SelectedContent, Mode=TwoWay}"
 						   behavior:TreeViewBehavior.AutoToggleItemExpansion="True"
+						   behavior:TreeViewBehavior.DoubleClickCommand="{Binding DoubleClickCommand}"
```
Xiaoy312, line 790: Double-clicking can trigger unintentionally. Wouldn't it make more sense to put this in a r-click menu with "save as ..."?
image

Uno/NugetPackageExplorer.Legacy/Views/Behaviors/TreeViewBehavior.cs
```diff
@@ -168,5 +194,43 @@ void ToggleExpansion(TreeView sender, TreeViewItemInvokedEventArgs args)
...
+
+        [SuppressMessage("Reliability", "CA2000:Dispose objects before losing scope", Justification = "managed via DoubleClickCommandDisposable")]
+        private static void OnDoubleClickCommandChanged(TreeView control, DependencyPropertyChangedEventArgs e)
+        {
+            GetDoubleClickCommandDisposable(control)?.Dispose();
+            if (e.NewValue is ICommand command)
+            {
+                var subscriptions = new CompositeDisposable();
+
+                var doubleTapped = Observable
```
Xiaoy312, line 206: For L204~234, you can skip the publish + connect, and the CompositeDisposable. They are used in [ListViewExtensions.cs#L115](https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/blob/4e33207ee9517ecccf1fccd839ab2bab9c21917e/Uno/NugetPackageExplorer.Legacy/Views/Extensions/ListViewExtensions.cs#L115) to avoid double subscriptions. But, it is not needed here.

```cs
var subscription = Observable
    .FromEventPattern<>(...)
    .Select(...)
    .Subscribe(...);
SetDoubleClickCommandDisposable(control, subscription);
```

Uno/NugetPackageExplorer.Legacy/Presentation/Content/InspectPackageViewModel.cs
```diff
@@ -268,6 +271,32 @@ public static async Task<InspectPackageViewModel> CreateFromRemotePackageWithFal
             }
         }
 
+        public async Task DoubleClick()
+        {
+            if (SelectedContent is IFile file)
```
Xiaoy312, line 276: I would wrap this entire method in `#if __WASM__ || __MACOS__` or `#if !HAS_UNO_SKIA_GTK`, [because FileSavePicker arnt supported yet for GTK](https://github.com/unoplatform/uno/blob/master/doc/articles/features/windows-storage-pickers.md). While no one uses NPE on that platform atm, but it is better to be safe.

campersau: I guess it will be supported soon: https://github.com/unoplatform/uno/pull/6544
But yes, I could #if it for now.

Uno/NuGetPackageExplorer.Shared/InspectPackage.xaml
campersau: Yes, that it also possible. But right clicks to open custom context menus aren't very common on the web, that's why I went for  double click. (It is also used to download a package from the feed, so there is a little symmetry to that.)

Xiaoy312 (approved):

PR #1360, opened 2021-09-03
Avoid crash when NuGet.Config file is not valid and show error to user

campersau:
When the `NuGet.Config` file contains invalid XML it would result in a crash when opening the package chooser window.

Now it willl show the error to the user:
image

We already did that for publishing:
https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/blob/b7b35c2c0100a6cc249b4ddc7b8501c316fc8a2b/PackageViewModel/PackageViewModel.cs#L907


clairernovotny:
Can we make this one better by showing a dialog with a "reset to default" button? Not sure if that should simply delete the file and let NuGet recreate it or if there's a default NuGet.config file that should be written instead.

clairernovotny:
Ping @campersau?

mudanzasIPROM (approved):

PR #1422, opened 2021-12-06
chore: Update to Uno 4.0

jeromelaban:
This PR updates the Uno Platform dependencies after the breaking changes introduced in 4.0.


clairernovotny:
Does Uno support targeting net6.0 now? I.e., can we nix the net5.0 support we have?

jeromelaban:
Yes, let me add that to the PR.

PR #1540, opened 2022-09-24
[Uno] adjustments based on feedback

lukeblevins:
_Discussed internally with @jeromelaban_


lukeblevins:
@jeromelaban I think we're good to go with these adjustments

Uno/NuGetPackageExplorer.Shared/Legacy/Presentation/ShellPageViewModel.cs
```diff
@@ -0,0 +1,35 @@
...
+        }
+
+        public ShellPageViewModel()
+        {
+
+#if __WASM__
+            IsStandaloneWindow = WebAssemblyRuntime.InvokeJS(@"window.matchMedia('(display-mode: standalone)').matches") == "true";
+#else
+            IsStandaloneWindow = true;
+#endif
```
jeromelaban, line 32: I wonder if we should be evaluating this on every window size change (when moving from the main window to PWA)

lukeblevins: @jeromelaban It didn't appear significant towards the user experience to do things that way. I also noticed invoking JS that many times seemed inefficient. In my view, we can iterate on that part later but it seems like a good idea to move forward with these changes now.

jeromelaban:
@lukeblevins It would be best if you were to split this PR in multiple ones, there are too many changes in there to make it easy to review.

clairernovotny:
@jeromelaban LGTM whenever you're ready to merge

PackageExplorer/Converters/ReferenceSetConverter.cs
```diff
@@ -19,16 +19,22 @@ public class ReferenceSetConverter : IValueConverter
...
+                }
 
-            // otherwise, flatten the groups into a single list of dependencies
-            return referenceSets.SelectMany(d => d.References);
+                // otherwise, flatten the groups into a single list of dependencies
+                return referenceSets.SelectMany(d => d.References);
+            }
+            else
+            {
+                return new List<PackageReferenceSet>();
```
jeromelaban, line 36: As we are returning an enumerable, let's not allocate further with a 
```suggestion
                return Enumerable.Empty<PackageReferenceSet>();
```

Uno/NuGetPackageExplorer.Shared/Dialogs/DownloadProgressDialog.xaml
```diff
@@ -1,45 +1,48 @@
...
+        <StackPanel Grid.Row="0"
+                    Spacing="4"
+                    MinWidth="256">
+            <TextBlock Style="{StaticResource TypoTitle1}" Text="Downloading" />
+            <!-- <muxc:AnimatedVisualPlayer x:Name="player"
+									   Width="256"
+									   Height="256"
+                                       AutoPlay="True">
+                <lottie:LottieVisualSource UriSource="ms-appx:///Assets/loading.json" />
+            </muxc:AnimatedVisualPlayer> -->
```
jeromelaban, line 26: Looks like this was added then removed. In most cases anyways until we get threading, animations won't run (even when based on lottie).

Uno/NuGetPackageExplorer.Shared/Legacy/Views/Behaviors/TreeViewBehavior.cs
```diff
@@ -195,35 +207,18 @@ void ToggleExpansion(TreeView sender, TreeViewItemInvokedEventArgs args)
...
-                        {
-                            typeof(TreeViewBehavior).Log().Error("failed to execute DoubleClick command: ", ex);
-                        }
-                    });
+        #endregion
+        #region DependencyProperty: AutoToggleItemExpansionDisposable
 
-                SetDoubleClickCommandDisposable(control, subscription);
-            }
-        }
```
jeromelaban, line 227: It's best to avoid cosmetic changes along with behavior changes, it makes reviewing more difficult.

lukeblevins:
@jeromelaban I've grouped these changes into separate commits and reset the breaking solution file changes. The major change is design, yet I moved a large folder around (apologies)

PR #1554, opened 2022-10-28
fix(wasm): Allow for packages with metadata in the version to be downloaded properly

jeromelaban:
Fixes the ability for packages to be downloaded in the wasm version, when version contains metadata (1.2.5+sha123abdef).


PackageExplorer/MefServices/PackageDownloader.cs
```diff
@@ -177,14 +177,20 @@ public async Task Download(string targetFilePath, SourceRepository sourceReposit
             {
                 try
                 {
+                    Console.WriteLine($"Downloading1: {packageIdentity.Id}");
```
clairernovotny, line 180: Should these be Console writes or use an ILogger, etc?

jeromelaban: They should not be there at all! Bad fixup, will fix. Thanks!

PR #1589, opened 2023-02-16
Remove unused `args` argument

Youssef1313:
https://github.com/unoplatform/uno/blob/45c85c08ca4e52dbc62fb11feb19b26996e15bbb/src/Uno.UI.Runtime.Skia.Gtk/GtkHost.cs#L67-L79

The constructor that takes `args` will be removed in Uno 5 and `args` is already unused.


jeromelaban:
Will be addressed in a separate PR.

PR #1627, opened 2023-09-30
Support more NuGet feeds for symbol validation than just nuget.org.

### LINKED ISSUE #1626: Add support to validate symbols from online feeds other than just nuget.org
# Feature request

It is not uncommon to publish packages to other feeds than nuget.org. In such cases, NPE is currently unable to to check, whether symbols where uploaded or not (health check).

I prepared a PR, that can support other feeds as well.

lauxjpn:
Feeds can differ in how the package ID and version has to be formatted.
This implementation supports nuget.org and myget.org (and possibly others).

Fixes #1626

### DIFF
--- a/Core/SymbolValidation/SymbolValidator.cs
+++ b/Core/SymbolValidation/SymbolValidator.cs
@@ -21,18 +21,20 @@ namespace NuGetPe
     {
         private readonly IPackage _package;
         private readonly string _packagePath;
+        private readonly string _packageSource;
         private readonly IFolder _rootFolder;
         private readonly HttpClient _httpClient;
 
-        public SymbolValidator(IPackage package, string packagePath, IFolder? rootFolder = null)
-            : this(package, packagePath, rootFolder, httpClient: null)
+        public SymbolValidator(IPackage package, string packagePath, string packageSource, IFolder? rootFolder = null)
+            : this(package, packagePath, packageSource, rootFolder, httpClient: null)
         {
         }
 
-        public SymbolValidator(IPackage package, string packagePath, IFolder? rootFolder, HttpClient? httpClient)
+        public SymbolValidator(IPackage package, string packagePath, string packageSource, IFolder? rootFolder, HttpClient? httpClient)
         {
             _package = package ?? throw new ArgumentNullException(nameof(package));
             _packagePath = packagePath ?? throw new ArgumentNullException(nameof(packagePath));
+            _packageSource = packageSource;
             _rootFolder = rootFolder ?? PathToTreeConverter.Convert(package.GetFiles().ToList());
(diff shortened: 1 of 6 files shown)


Core/SymbolValidation/SymbolValidator.cs
```diff
@@ -46,15 +48,7 @@ public async Task<SymbolValidatorResult> Validate(CancellationToken cancellation
         {
             try
             {
-                // NuGet signs all its packages and stamps on the service index. Look for that.
-                if (_package is ISignaturePackage sigPackage)
-                {
-                    await sigPackage.LoadSignatureDataAsync().ConfigureAwait(false);
```
clairernovotny, line 52: Why is the signature load removed -- I didn't see it called elsewhere?

Core/SymbolValidation/SymbolValidator.cs
```diff
@@ -269,24 +263,52 @@ private async Task<SymbolValidatorResult> CalculateValidity(IReadOnlyList<IFile>
...
-                        if (response.IsSuccessStatusCode) // we'll get a 404 if none
+                        if (symbolPackagePublishSourceUri is not null)
                         {
+                            var variants = new[]
+                            {
+                                async () => await ReadPublicPackageSymbolFile(_package.Id.ToLowerInvariant(), _package.Version.ToNormalizedString().ToLowerInvariant()), // nuget.org
+                                async () => await ReadPublicPackageSymbolFile(_package.Id, _package.Version.OriginalVersion), // myget.org
+                            };
+
+                            foreach (var variant in variants)
```
clairernovotny, line 277: Can we be more intelligent here so we're not blindly trying requests?

clairernovotny (changes requested):
Thank you -- overall looks great, just needs a little cleanup

Core/SymbolValidation/SymbolValidator.cs
lauxjpn: It has been too long for me to remember. I am sure I thoroughly debugged the code and there was a good reason for it (e.g. MyGet would not work otherwise. Do you want me to check/debug again why this change was necessary?

Core/SymbolValidation/SymbolValidator.cs
lauxjpn: Would be nice, but I couldn't figure it out in the short time I spend on this feature, so that I could check that our nightly build packages were properly configured in regards to symbol support.

lauxjpn: Feel free to directly commit changes to the PR until it reaches the quality necessary to be merged, if you want to (probably quicker).

PR #1643, opened 2023-12-07
Fix displaying multiple AssemblyMetadata attribute values

akoeplinger:
We were using a Dictionary to store the assembly metadata with the attribute name as the key, but an assembly can have multiple instances of an attribute e.g. the `AssemblyMetadataAttribute` so we were overwriting the previous values.

**Before:**

image

<p></p>

**After:**

image


jeromelaban (approved):
Validated on web as well, look good to me!

PR #1675, opened 2024-08-10
Add ITemporaryPathProvider to allow more temp file customization

joelverhagen:
My project (NuGet Insights) is getting alerts for some temp files generated by SymbolValidator. I can tell because ".npe" is seen in the file extension, per https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/pull/1670. I would like a way of fully customizing the temporary files so I can enhance my logging, select specific temporary directory, etc.

### DIFF
--- a/Core/SymbolValidation/SymbolValidator.cs
+++ b/Core/SymbolValidation/SymbolValidator.cs
@@ -205,13 +205,12 @@ namespace NuGetPe
                     try
                     {
 
-                        using var str = file.Primary.GetStream();
+                        using var stream = file.Primary.GetStream();
 
                         // Use descriptive file extension so files that appear in file system logging or that are
                         // leftover during an abrupt process termination can be debugged easier
                         var tempFileExtension = ".npe" + (string.IsNullOrEmpty(file.Primary.Extension) ? ".dat" : file.Primary.Extension);
-
-                        using var tempFile = GetTemporaryFile(str, tempFileExtension, file.Primary.Name);
+                        using var tempFile = GetTemporaryFile(stream, extension: tempFileExtension, part: file.Primary);
 
                         var assemblyMetadata = AssemblyMetadataReader.ReadMetaData(tempFile.FileName);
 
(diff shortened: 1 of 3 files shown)


Core/SymbolValidation/SymbolValidator.cs
```diff
@@ -290,7 +297,7 @@ private async Task<SymbolValidatorResult> CalculateValidity(IReadOnlyList<IFile>
 #else
                             using var getStream = await response.Content!.ReadAsStreamAsync().ConfigureAwait(false);
 #endif
-                            using var tempFile = new TemporaryFile(getStream, ".npe.snupkg");
+                            using var tempFile = GetTemporaryFile(getStream, extension: ".npe.snupkg", name: fileName);
```
clairernovotny, line 300: Is there any reason to put `.npe.` in the filename instead of just using the original file name?

Core/SymbolValidation/SymbolValidator.cs
```diff
@@ -23,18 +23,25 @@ public class SymbolValidator
...
             : this(package, packagePath, rootFolder, httpClient: null)
         {
         }
 
         public SymbolValidator(IPackage package, string packagePath, IFolder? rootFolder, HttpClient? httpClient)
+            : this(package, packagePath, rootFolder, httpClient, tempProvider: null)
+        {
+        }
+
+        public SymbolValidator(IPackage package, string packagePath, IFolder? rootFolder, HttpClient? httpClient, ITemporaryFileProvider? tempProvider)
```
clairernovotny, line 38: Would it be better to have a default implementation of `ITemporaryFileProvider` registered in DI instead of null checks and if statements later?

Core/SymbolValidation/SymbolValidator.cs
joelverhagen: I didn't want to run until long path issues. `.npe.nupkg` is less likely than `{long ID}.{long version}.snupkg`. Additionally, the .snupkg temp file was never triggering the alerts. It was the assembly temp file case.

Core/SymbolValidation/SymbolValidator.cs
joelverhagen: I didn't want to make assumptions on what the temp file provider would do and instead leave the fully up to the user. Maybe if we get feedback from folks using the Core package on common usage patterns?

PR #1692, opened 2024-11-29
Bump to net9

jeromelaban:
- Bumps all appropriate projects to net9
- Bumps uno to latest


martincostello:
I would suggest keeping the net8.0 TFM and instead multi-targeting - otherwise people using .NET 8 LTS and not using .NET 9 will need to install .NET 9 just to run the tool.

jeromelaban:
> I would suggest keeping the net8.0 TFM and instead multi-targeting for `dotnet-validate` - otherwise people using .NET 8 LTS and not using .NET 9 will need to install .NET 9 just to run the tool.

That's fair. It was not done in the past, that's why I kept it this way, but I'll do the change.

Uno/NuGetPackageExplorer/Platforms/WebAssembly/wwwroot/staticwebapp.config.json
```diff
@@ -8,6 +8,13 @@
     ]
   },
   "routes": [
+    {
+      "route": "/_framework/*",
+      "headers": {
+        "cache-control": "public, immutable, max-age=31536000"
```
clairernovotny, line 14: Will this cause any caching issues or are the files deployed with unique path/querstring args?

jeromelaban: I don't think so. These files are never referenced as part of query strings, and are equivalent to the `/package_*` route below. This folder contains the dlls now (in a `.wasm` format).

clairernovotny (approved):

